PR #9, opened 2021-05-04
first solution

dvdovynets:


src/test/java/core/basesyntax/model/FruitTest.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.model;
+
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class FruitTest {
```
freedmanMain, line 7: You don't need to test classes that have no functionality. It's just a model.

src/test/java/core/basesyntax/dto/TransactionDtoTest.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.dto;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.Operation;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class TransactionDtoTest {
```
freedmanMain, line 9: There is no point in testing this class. The DTO has no functionality whatsoever. It can only store data about an entity and you can read it out.

freedmanMain (approved):

src/test/java/core/basesyntax/model/FruitTest.java
dvdovynets: If I would made those tests after morning lecture I would add Fruit and TransactionDto together with main to pom.xml. But the pluses are I know that all get methods and equals works correctly.

src/test/java/core/basesyntax/services/impl/FileWriterServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void beforeClass() {
+        fileWriterService = new FileWriterServiceImpl();
+    }
+
+    @Test
+    public void write_CorrectReportAndFilePath() {
+        fileWriterService.write(CORRECT_REPORT, CORRECT_FILE_PATH);
+        List<String> expected;
+        try {
+            expected = Files.readAllLines(Path.of(CORRECT_TEST_REPORT));
```
RomanDubovskyi, line 32: if you write correct report to file why not use variable CORRECT_REPORT as expected ?

src/test/java/core/basesyntax/services/impl/FileReaderServiceImplTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @BeforeClass
+    public static void beforeClass() {
+        fileReaderService = new FileReaderServiceImpl();
+    }
+
+    @Test
+    public void read_TestCorrectInput() {
+        List<String> expected;
+        try {
+            expected = Files.readAllLines(Path.of(CORRECT_FILE_PATH));
```
RomanDubovskyi, line 26: same here you can just use constant from FileWriterServiceImplTest

src/test/java/core/basesyntax/model/FruitTest.java
RomanDubovskyi, line 7: we don't tests model classes unless they have some additional buisness logic

src/test/java/core/basesyntax/dto/TransactionDtoTest.java
RomanDubovskyi, line 9: same here

RomanDubovskyi (changes requested):

src/test/java/core/basesyntax/dto/TransactionDtoTest.java
dvdovynets: Cleared

src/test/java/core/basesyntax/model/FruitTest.java
dvdovynets: Cleared

src/test/java/core/basesyntax/services/impl/FileWriterServiceImplTest.java
dvdovynets: Changed expected to a predefined List<Strings>

src/test/java/core/basesyntax/services/impl/FileReaderServiceImplTest.java
dvdovynets: Did the same here

RomanDubovskyi (approved):

PR #25, opened 2021-05-05
first solution

andriiromaniuk:


IhorRome (approved):

src/test/java/service/ActivityStrategyTest.java
```diff
@@ -0,0 +1,48 @@
...
+        dataValidator = new DataValidatorImpl();
+        fruitDao = new FruitDaoImpl();
+    }
+
+    @Test
+    public void get_putDataAndGetIt_OK() {
+        Map<String, ActivityHandler> data = new HashMap<>();
+        data.put("BalanceHandler", new Balance(fruitDao));
+        data.put("PurchaseHandler", new Purchase(fruitDao, dataValidator));
+        data.put("SupplyOrReturn", new SupplyOrReturn(fruitDao));
```
Max-Pochepets, line 33: we can do this in `BeforeClass`

src/test/java/service/file/FileWriterTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Test
+    public void writeToFile_GoodPath_Ok() {
+        File file = new File(GOOD_PATH);
+        fileWriter.writeToFile(file.getPath(), GOOD_DATA);
+        List<String> expected = EXPECTED_LIST;
+        try {
+            List<String> actual = Files.readAllLines(file.toPath());
+            assertEquals(expected, actual);
+            Files.write(file.toPath(), "".getBytes());
```
Max-Pochepets, line 38: why do we need to write afterwards?

Max-Pochepets (changes requested):
good job 👍

src/test/java/service/file/FileWriterTest.java
andriiromaniuk: Cleaning file, don't need data in it afterwards anyway. Will delete this for now, don't know where else can i put it or if i need it.

Max-Pochepets: Ok. if you you have only one file - it's fine. But you can do this in `AfterClass` in this situation, or `After`, depends on circumstances.

Max-Pochepets (approved):
LGTM

src/test/java/service/file/FileWriterTest.java
andriiromaniuk: okay, will remember, thanks. : )

PR #87, opened 2021-09-21
Fruit shop tests hw

AlexPravda1:
- Exceptions tested
- no tests for Main Class
- 80+ coverage


src/test/java/core/basesyntax/ReportServiceImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Before
+    public void setUp() {
+        activityHandlerMap.put("b", balanceActivityHandler);
+        activityHandlerMap.put("p", purchaseActivityHandler);
+        activityHandlerMap.put("r", returnActivityHandler);
+        activityHandlerMap.put("s", supplyActivityHandler);
+    }
+
+    @Test
+    public void reportServiceImplConstructor_Ok() {
```
RomanDubovskyi, line 51: this check will never fail, I think, we can remove it

src/test/java/core/basesyntax/InputValidatorImplTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import core.basesyntax.service.InputValidator;
+import core.basesyntax.service.InputValidatorImpl;
+import java.util.List;
+import org.junit.Test;
+
+public class InputValidatorImplTest {
+    private static final String HEADER = "type,fruit,quantity";
+    private static final List<String> listOk = List.of(HEADER, "s,banana,100");
+    private static final List<String> listNotOk1 = List.of(HEADER, ",banana,100");
```
RomanDubovskyi, line 13: we can improve naming here :
```suggestion
    private static final List<String> invaliOperationList = List.of(HEADER, ",banana,100");
```
same for other input names

src/test/java/core/basesyntax/CsvFileServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class CsvFileServiceImplTest {
+    private static final String OK_READ_FILE_PATH = "src/test/resources/CsvFileTestInput.csv";
+    private static final String OK_WRITE_FILE_PATH = "src/test/resources/FileWriteDaoTest.csv";
+    private static final String NOT_OK_FILE_PATH = "src/test/no such path";
+    private static final String STRING_TO_WRITE = "apple, banana";
+    private static final List<String> EXPECTED_LIST = List.of("apple", "banana");
```
RomanDubovskyi, line 16: let's move this to the test method where it's used, expected result is better to declare in the methods

RomanDubovskyi (changes requested):

src/test/java/core/basesyntax/CsvFileServiceImplTest.java
AlexPravda1: Shall I move all 4 variables? I've relocated List EXPECTED_LIST since I'm using it only once. The rest of the Strings are used multiple times within the tests, shall I declare them in each relevant test?

src/test/java/core/basesyntax/ReportServiceImplTest.java
RomanDubovskyi: if constructor doesn't have any custom logic we can ignore test for it.

src/test/java/core/basesyntax/CsvFileServiceImplTest.java
RomanDubovskyi: That was only concerning EXPECTED_LIST, it's ok to have generic variables in class that may be reused and are not test case specific

RomanDubovskyi (approved):

PR #133, opened 2021-11-21
HW Nishchenko

nishchenkokateryna:


src/test/java/core/basesyntax/service/ShopServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ShopServiceTest {
+    private static ShopService shopService;
+    private static final Map<String, ActivityHandler> activityHandlerMap = new HashMap<>();
+    private static List<String> INPUT_DATA;
```
DmitroKoz, line 25: Is this constant?

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ValidatorTest {
+    private static Validator validator;
+    private static final Map<String, ActivityHandler> activityHandlerMap = new HashMap<>();
```
DmitroKoz, line 20: hm... Do we need to declare Map here?

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    private List<String> inputData;
+
+    @BeforeClass
+    public static void beforeClass() throws Exception {
```
DmitroKoz, line 24: What will happen if we will delete this throws?

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void validateData_validData_ok() {
+        inputData = List.of("type,fruit,quantity",
+                "b,banana,20",
+                "b,apple,100",
+                "s,banana,100",
+                "p,banana,13");
+        boolean actual = validator.validateData(inputData);
+        Assert.assertTrue(actual);
+    }
+}
```
DmitroKoz, line 102: What will happen if we will get such input data?
0100
12.5
12,5

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+        inputData = List.of("type,fruit,quantity",
+                "b,banana,-20",
+                "b,apple,-100",
+                "s,banana,-100",
+                "p,banana,-13");
+        validator.validateData(inputData);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void validateData_dataWithWrongActivityType_notOk() {
```
DmitroKoz, line 63: What will happen if we will use uppercase S letter as activityType?
Doubled "ss"?

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,52 @@
+package core.basesyntax;
```
DmitroKoz, line 1: How about to use @Rule to make tests for exceptions? You use different exception messages in your code so mey be its better to check it?

Not sure that is mistake but...
In many cases:
        inputData = List.of("type,fruit,quantity",
                "b,1banana,20",
                "b,2apple,100",
                "s,44444banana,100",
                " p,banana,13");
What will happen if your class will throw exception at  "b,1banana,20" lina but will pass "s,44444banana,100"?
Isn't better to use always right List and set just 1 disright line to it?

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test(expected = RuntimeException.class)
+    public void validateData_nullData_notOk() {
+        validator.validateData(null);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void validateData_emptyData_notOk() {
+        validator.validateData(Collections.emptyList());
+    }
+
```
DmitroKoz, line 42: What will happen if your input file contains just 1 line (header).

src/test/java/core/basesyntax/service/ValidatorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test(expected = RuntimeException.class)
+    public void validateData_dataWithWrongFruitName_notOk() {
+        inputData = List.of("type,fruit,quantity",
+                "b,1banana,20",
+                "b,2apple,100",
+                "s,44444banana,100",
+                " p,banana,13");
+        validator.validateData(inputData);
+    }
+
```
DmitroKoz, line 81: What will happen if your fruitName will contains uppercase letters?

src/test/java/core/basesyntax/service/ValidatorTest.java
DmitroKoz, line 42: What will happen if header will be incorrect?

DmitroKoz, line 42: What will happen if we will not have header?

src/test/java/core/basesyntax/service/WriterTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test(expected = RuntimeException.class)
+    public void writeToFile_validDataAndNullPath_notOk() {
+        writer.writeToFile(VALID_REPORT, null);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void writeToFile_nullDataAndNullPath_notOk() {
+        writer.writeToFile(null, null);
+    }
```
DmitroKoz, line 39: Is that method notOk cause of data is null or cause of Path is null?

src/test/java/core/basesyntax/service/ShopServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+                "p,banana,5",
+                "s,banana,2");
+        shopService.updateShopWarehouse(INPUT_DATA);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void updateShopWarehouse_nullData_notOk() {
+        shopService.updateShopWarehouse(null);
+    }
+}
```
DmitroKoz, line 84: What will happen if our 1st line will be "p,apple,1"?

DmitroKoz, line 84: What will happen if one of line will be "p,apple,1000000000000"

DmitroKoz: What will happen if our quantity (especially purchase) will be bigger than Integer_MaxValue?

DmitroKoz (changes requested):

src/test/java/core/basesyntax/service/WriterTest.java
nishchenkokateryna: Both causes - null data and null path

src/main/java/core/basesyntax/Main.java
nishchenkokateryna: About @Rule - thanks for advice

DmitroKoz (approved):

KseniiaMakarova (changes requested):
overall great job! let's also add tests for handlers

KseniiaMakarova (approved):

PR #194, opened 2022-02-18
test first ver

LinkClink:


DzhuzhaMichael (approved):

DzhuzhaMichael:
Looks good!

src/test/java/core/basesyntax/service/impl/FileWriteServiceImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @BeforeClass
+    public static void setUp() {
+        goodData = "fruit,quality" + System.lineSeparator()
+                + "banana,152" + System.lineSeparator()
+                + "apple,90";
+        goodFilePath = "src/test/resources/out.csv";
+    }
+
+    @Test
+    public void fileWrite_goodFilePath_Ok() throws IOException {
```
KostyaNikiforov, line 26:
```suggestion
    public void fileWrite_goodFilePath_Ok() {
```
it is not necessary

src/test/java/core/basesyntax/service/impl/GenerateReportDataImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void generateReport_emptyData_Ok() {
+        String actual = generateReportDataService.report(data);
+        String expected = "fruit,quality"
+                + System.lineSeparator();
+        assertEquals(expected, actual);
+    }
```
KostyaNikiforov, line 32: It would be cool to add here at least one negative test case

src/test/java/core/basesyntax/service/impl/WriteDataToStorageImplTest.java
```diff
@@ -0,0 +1,73 @@
...
+import core.basesyntax.service.file.WriteDataToStorageService;
+import org.junit.After;
+import org.junit.Test;
+
+public class WriteDataToStorageImplTest {
+    private WriteDataToStorageService writeDataToStorageService = new WriteDataToStorageImpl();
+    private TransactionStorage transactionStorage = new TransactionStorage();
+
+    @Test
+    public void writeDataToStorage_firstLine_Ok() {
```
KostyaNikiforov, line 15: First part of test method name should be name of the method you test 
```suggestion
    public void writeData_firstLine_Ok() {
```

KostyaNikiforov (changes requested):
Let's add everywhere a test case when the input is null

KostyaNikiforov (approved):

src/main/java/core/basesyntax/service/impl/FileWriteServiceImpl.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class FileWriteServiceImpl implements FileWriteService {
+
+    @Override
+    public void writeDataToFile(String data, String filePath) {
+        try (BufferedWriter bufferedWriter
+                     = new BufferedWriter(new FileWriter(filePath, false))) {
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file" + e);
```
liliia-ponomarenko, line 16:
```suggestion
            throw new RuntimeException("Can't write data to file" + e);
```

PR #199, opened 2022-02-21
Hw solution

Sasha-Bachynskyi:


src/test/java/fruite/store/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void addFruitToStorage_quantityLessThanZero_notOk() {
+        storageDao.addFruitToStorage("apple", -1);
+    }
+
+    @Test
+    public void addFruitToStorage_validData_ok() {
+        int sizeBeforeAddingData = Storage.fruitStorage.size();
```
LinkClink, line 25: Maybe name (expected) instead of  `sizeBeforeAddingData`

src/test/java/fruite/store/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+        storageDao.addFruitToStorage("apple", 50);
```
LinkClink, line 26: Maybe name (actual) instead of  `sizeAfterAddingData`

src/test/java/fruite/store/service/impl/FileWriterServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void afterClass() {
+        try {
+            Files.deleteIfExists(Path.of(VALID_FILE_NAME_AND_PATH));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't correctly clear result files after test: ", e);
+        }
+    }
+
+    @Test
+    public void writeToFile_validPathOrFileName_ok() {
```
LinkClink, line 29: You can compare with `String `  using readString to get file data ( to simplify this)

LinkClink (approved):
Look this

KostyaNikiforov (changes requested):
Let's add everywhere a test case when the input is null to be confident that it throws RE and not NPE

KostyaNikiforov (approved):

PR #203, opened 2022-02-21
Solution was covered with tests with 96% coverage(excluding Main class).

karma-o:
minor adjustments of the original fruit shop was made during tests creation.

### DIFF
--- a/src/test/java/core/basesyntax/dao/FileReaderTest.java
+++ b/src/test/java/core/basesyntax/dao/FileReaderTest.java
@@ -1,0 +1,47 @@
+package core.basesyntax.dao;
+
+import static org.junit.Assert.assertEquals;
+
+import core.basesyntax.dao.impl.CsvReaderImpl;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class FileReaderTest {
+    public static final String OK_TEST_PATH = "./src/test/java/resources/test_Ok.csv";
+    private static FileReader fileReader;
+
+    @BeforeClass
+    public static void beforeClass() {
+        fileReader = new CsvReaderImpl();
+    }
+
+    @Test
+    public void csvReader_parse_Ok() {
+        List<String> expected = new ArrayList<>();
+        expected.add("type,fruit,quantity");
+        expected.add("b,banana,20");
+        expected.add("b,apple,100");
+        expected.add("s,banana,100");
+        expected.add("p,banana,13");
+        expected.add("r,apple,10");
+        expected.add("p,apple,20");
+        expected.add("p,banana,5");
+        expected.add("s,banana,50");
+        List<String> actual = fileReader.parse(OK_TEST_PATH);
+        assertEquals(expected, actual);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void csvReader_parse_fileDoesNotExist_notOk() {
+        fileReader.parse("non existent file");
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void csvReader_parse_emptyFile_notOk() {
+        File testInput = new File("./src/test/java/resources/test_empty.csv");
+        fileReader.parse("./src/test/java/resources/test_empty.csv");
+    }
+}
--- a/src/test/java/core/basesyntax/dao/StorageDaoTest.java
+++ b/src/test/java/core/basesyntax/dao/StorageDaoTest.java
@@ -1,0 +1,82 @@
+package core.basesyntax.dao;
+
+import static core.basesyntax.db.Storage.storage;
+import static org.junit.Assert.assertEquals;
+
+import core.basesyntax.dao.impl.StorageDaoImpl;
+import core.basesyntax.model.Fruit;
+import java.util.NoSuchElementException;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class StorageDaoTest {
+    private static StorageDao storageDao;
+    private Fruit defaultFruit;
+
+    @BeforeClass
+    public static void beforeClass() {
+        storageDao = new StorageDaoImpl();
+    }
+
+    @Before
+    public void setUp() {
+        defaultFruit = new Fruit("banana", 100);
+    }
+
+    @Test
+    public void storageDao_add_Ok() {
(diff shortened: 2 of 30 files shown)


OlehMytiai:
It's better to create folder "resources" for implementation here -> "/src/main/resources" and for tests here -> "src/test/resources"

OlehMytiai:
Классы FileReader, ReportSupplier разве должны быть в пакете Dao? 
> В общем случае, определение [Data Access Object ](https://ru.wikipedia.org/wiki/Data_Access_Object)описывает его как прослойку между БД и системой.

OlehMytiai (approved):

src/test/java/core/basesyntax/dao/FileReaderTest.java
```diff
@@ -0,0 +1,55 @@
...
+        expected.add("b,apple,100");
+        expected.add("s,banana,100");
+        expected.add("p,banana,13");
+        expected.add("r,apple,10");
+        expected.add("p,apple,20");
+        expected.add("p,banana,5");
+        expected.add("s,banana,50");
+        List<String> actual = fileReader.parse(testFileOk.toString());
+        assertEquals(expected, actual);
+    }
```
KostyaNikiforov, line 39: It better to create test file with the data in a test class

src/test/java/core/basesyntax/dao/StorageDaoTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test
+    public void storageDao_add_Ok() {
+        storageDao.add("banana", 100);
+        Fruit expected = defaultFruit;
+        Fruit actual = storage.get(0);
+        //new fruit
+        assertEquals(expected.getName(), actual.getName());
+        assertEquals(expected.getAmount(), actual.getAmount());
+        //existing fruit
```
KostyaNikiforov, line 36: Uncle Bob says: don't explain with comments what code explains better)
Do we really need those comments?

KostyaNikiforov (changes requested):
Let's add negative test cases to classes:
OperationStrategyTest
ReportCreatorTest
DataProcessorTest

src/test/java/core/basesyntax/dao/FileReaderTest.java
karma-o: I am already using a test file in actual, expected is just a list with an expected result.
If you are talking about using a file in expected, I dont see the reason for that

src/test/java/core/basesyntax/dao/StorageDaoTest.java
karma-o: Forgot to remove them, will do

karma-o:
> Let's add negative test cases to classes: OperationStrategyTest ReportCreatorTest DataProcessorTest

I think I am already handling negative case or invalid amount in `StorageDao`, `substract` method

karma-o:
> Let's add negative test cases to classes: OperationStrategyTest ReportCreatorTest DataProcessorTest

I dont see any way or reason to write more tests to `OperationStrategyImpl`, this class has only 1 method `get()` which takes an enum. We wont be able to put anything else as a parameter to this method, so all we can do is test the proper work with valid data in our `operationStrategy_get_Ok()` test. Also I dont see any other way to break it and produce a suitable exception.

About `ReportCreatorService` tests, situation is basically the same: storage is either empty or has some elements and takes no parameters to create a report, it takes data diractly from `StorageDao` which is passed through constructor, but just to be sure I have writen 1 more test: `reportCreatorService_createReport_emptyStorage_Ok()`

About `DataProcessor`, added couple new tests as you pointed out:
`dataProcessor_registerTransaction_noSuchElement_notOk`
`dataProcessor_registerTransaction_invalidAmount_notOk`

src/main/java/core/basesyntax/service/impl/CsvReaderImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import core.basesyntax.service.FileReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class CsvReaderImpl implements FileReaderService {
+
+    @Override
+    public List<String> parse(String filePath) {
```
KostyaNikiforov, line 12:
```suggestion
    public List<String> read(String filePath) {
```

src/test/java/core/basesyntax/service/FileReaderTest.java
```diff
@@ -0,0 +1,54 @@
...
+        fileReader.parse(null);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void csvReader_parse_fileDoesNotExist_notOk() {
+        fileReader.parse("non existent file");
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void csvReader_parse_emptyFile_notOk() {
```
KostyaNikiforov, line 51:
```suggestion
    public void parse_emptyFile_notOk() {
```
The first part of the test method name must be the name of the method under test.

src/test/java/core/basesyntax/service/FileWriterTest.java
```diff
@@ -0,0 +1,85 @@
...
+        testReportFile = new File("./src/resources/test_report.csv");
+    }
+
+    @Before
+    public void setUp() {
+        testReportFile.delete();
+    }
+
+    @Test
+    public void fileWriter_writeStringToFile_Ok() {
```
KostyaNikiforov, line 37:
```suggestion
    public void writeStringToFile_validData_ok() {
```

src/test/java/core/basesyntax/service/ReportCreatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+        String expectedBanana = "banana,100";
+        String expectedOrange = "orange,10";
+        String actualBanana = reportCreatorService.createReport().get(0);
+        String actualOrange = reportCreatorService.createReport().get(1);
+        assertEquals(expectedBanana, actualBanana);
+        assertEquals(expectedOrange, actualOrange);
+    }
+
+    @Test
+    public void reportCreatorService_createReport_emptyStorage_Ok() {
```
KostyaNikiforov, line 35:
```suggestion
    public void createReport_emptyStorage_ok() {
```

KostyaNikiforov (changes requested):

PR #287, opened 2022-06-16
Fruit shop tests added

antonEEmus:


src/test/java/core/basesyntax/ProcessorServiceTest.java
```diff
@@ -0,0 +1,70 @@
...
+                "r,grapefruit,2"));
+        Assert.assertEquals(dao.get("apple"), 65);
+        Assert.assertEquals(dao.get("banana"), 62);
+        Assert.assertEquals(dao.get("orange"), 6);
+        Assert.assertEquals(dao.get("grapefruit"), 1085);
+    }
+
+    @After
+    public void afterEach() {
+        Storage.fruitsAvailable.clear();
```
zim10093, line 68: So, the main idea of unit testing is testing only one piece of code that can work separately. There I see that you use your actual storage from the main program. In general we need sometimes like the Mockito framework, but now we don't know about it. 
So create for test new Storage and use it by Dao

src/test/java/core/basesyntax/ReaderServiceTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void readNotExistingFile_shouldThrowRuntimeException_ok() {
+        try {
+            readerService.readFile(INVALID_FILE_NAME);
+        } catch (RuntimeException e) {
+            return;
+        }
+
+        Assert.fail("Should throw new RuntimeException");
+    }
```
zim10093, line 54: You can use @ Test(expected = RuntimeException.class) google it

src/test/java/core/basesyntax/ReportServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+                + "banana,62" + System.lineSeparator()
+                + "grapefruit,1085" + System.lineSeparator()
+                + "orange,6";
+        String actual = reportService.makeReport();;
+        Assert.assertEquals("expected: " + expected + ", actual: " + actual, expected, actual);
+    }
+
+    @After
+    public void afterEach() {
+        Storage.fruitsAvailable.clear();
```
zim10093, line 42: There the same. You use storage from main program

src/test/java/core/basesyntax/ReportServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+public class ReportServiceTest {
+    private static FruitDao dao;
+    private static ReportService reportService;
+
+    @BeforeClass
+    public static void beforeAll() {
+        dao = new FruitDaoImpl();
+        reportService = new ReportServiceImpl(dao);
```
zim10093, line 21:
```suggestion
        reportService = new ReportServiceImpl(new FruitDaoImp());

```
You need dao only to create reportService. I think this way to write is better

src/test/java/core/basesyntax/ReportServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+                "apple", 65,
+                "banana", 62,
+                "orange", 6,
+                "grapefruit", 1085));
+        String expected = "fruit,quantity" + System.lineSeparator()
+                + "apple,65" + System.lineSeparator()
+                + "banana,62" + System.lineSeparator()
+                + "grapefruit,1085" + System.lineSeparator()
+                + "orange,6";
+        String actual = reportService.makeReport();;
```
zim10093, line 36:
```suggestion
        String actual = reportService.makeReport();

```

src/test/java/core/basesyntax/WriteServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+                    "banana,62",
+                    "orange,1082",
+                    "grapefruit,6");
+            Assert.assertEquals(expected, actual);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file from path: " + FILE_NAME, e);
+        }
+    }
+
+    @Test
```
zim10093, line 45: You can use @ Test(expected = RuntimeException.class) too

src/test/java/core/basesyntax/WriteServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public void writeExistingFile_shouldThrowRuntimeException_ok() {
```
zim10093, line 46: If I understand correctly, this situation is "_notOk".
I seen it a few times.

zim10093 (changes requested):
I think you don`t check csv file if it is valid.
has the correct header.
has correct number of row
What happened if your program meet records like this "b, banana"?

src/test/java/core/basesyntax/ProcessorServiceTest.java
```diff
@@ -0,0 +1,82 @@
...
+        Assert.assertEquals(dao.get("apple"), 65);
+        Assert.assertEquals(dao.get("banana"), 62);
+        Assert.assertEquals(dao.get("orange"), 6);
+        Assert.assertEquals(dao.get("grapefruit"), 1085);
+    }
+
+    @Test
+    public void process_invalidHeader_ok() {
+        processorService.processData(List.of("aamcgoreimvleiotumovoci"));
+    }
```
zim10093, line 69: I see that you are ignoring the header.  So, maybe you need to check if the header is correct?
Because you can process the wrong table, bat it can be similar.

zim10093 (changes requested):

zim10093 (approved):

zim10093:
I think it better to check with an external header which is transmitted by arguments. but in this case, it may be

pom.xml
```diff
@@ -110,4 +110,4 @@
             </plugins>
         </pluginManagement>
     </build>
-</project>
+</project>
```
KostyaNikiforov, line 113: Add an empty line to the end of the file

src/main/java/core/basesyntax/dao/FruitDaoImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public Map<String, Integer> getAll() {
+        return storage;
+    }
+
+    @Override
+    public int get(String key) {
+        return storage.get(key) == null ? 0 : storage.get(key);
```
KostyaNikiforov, line 29: you can user method `containsKey()`

src/test/java/core/basesyntax/ReportServiceTest.java
```diff
@@ -0,0 +1,43 @@
...
+                "banana", 62,
+                "orange", 6,
+                "grapefruit", 1085));
+        String expected = "fruit,quantity" + System.lineSeparator()
+                + "apple,65" + System.lineSeparator()
+                + "banana,62" + System.lineSeparator()
+                + "grapefruit,1085" + System.lineSeparator()
+                + "orange,6";
+        String actual = reportService.makeReport();
+        Assert.assertEquals("expected: " + expected + ", actual: " + actual, expected, actual);
```
KostyaNikiforov, line 36: Isn't this message unnecessary? If values are not equal will see it in console in almost the same format

src/test/java/core/basesyntax/WriteServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+            Assert.assertEquals(expected, actual);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file from path: " + FILE_NAME, e);
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void writeExistingFile_shouldThrowRuntimeException_notOk() {
+        try {
+            Files.createFile(Path.of(FILE_NAME));
```
KostyaNikiforov, line 48: We don't need to create file here, because the `white` method does it automatically

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void writeFile(String data, String fileName) {
+        Path filePath = Path.of(fileName);
+        try {
+            Files.createFile(filePath);
```
KostyaNikiforov, line 14: It is redundant

src/test/java/core/basesyntax/WriteServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    @After
+    public void afterEach() {
+        try {
+            Files.delete(Path.of(FILE_NAME));
+        } catch (IOException e) {
+            return;
+        }
```
KostyaNikiforov, line 61: I guess you can add an exception to the method signature. Any way, you don't process it and if your tests work fine there shouldn't be a problem

KostyaNikiforov (changes requested):

liliia-ponomarenko (approved):
good job 👍

PR #289, opened 2022-06-17
tests were implemented

NikitaSemisoshenko:


src/test/java/core/basesyntax/dao/FruitsDaoImplTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void isFruitAvailable_checkForAvailability_ok() {
+        Assert.assertTrue(fruitsDao.isFruitAvailable("banana"));
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        Storage.fruits.clear();
```
antonEEmus, line 88: you should use not class Storage, but some dummy database for tests. it can be just map or class with map.

src/test/java/core/basesyntax/service/impl/CsvFileWriterImplTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class CsvFileWriterImplTest {
+    private static final String PATHNAME = "src/main/resources/report.csv";
+    private static final String PATHNAME_NOT_EXIST = "src/main/resources/reportNon.csv";
+    private static CsvFileWriter fileWriter;
+
+    private List<String> generatedListForWritingToFile;
+
+    @BeforeClass
+    public static void beforeClass() throws Exception {
```
antonEEmus, line 19: why are your beforeclasses throwing exception?

antonEEmus (changes requested):

PR #291, opened 2022-06-20
Homework fruit shop tests

Cosmic-M:


zim10093 (changes requested):
First of all, do like wrote in common mistakes file. After I will check that
"Create a corresponding test class for each service you test. Do not test logic of whole program in one test class. That’s important for code readability."

zim10093 (changes requested):

zim10093:
May it repeat twice because update page when writing comments

src/test/java/service/impl/ParseServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+        assertEquals("method parse didn't return expected quantity of objects",
+                2, parseService.parse(list).size());
+        assertEquals("method parse didn't return correspond result",
+                transaction1, parseService.parse(list).get(0));
+        assertEquals("method parse didn't return correspond result",
+                transaction2, parseService.parse(list).get(1));
+    }
+
+    @Test
+    public void parse_throwsRuntimeExceptionIfFetchedDataFormatWasIncorrect_notOk() {
```
zim10093, line 41: I think the name is too long. use like this " parse_ invalidFormat_notOk"

src/test/java/service/impl/ParseServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+        List<String[]> list = new ArrayList<>();
+        list.add(new String[]{"33", "cows"});
+        exception.expect(RuntimeException.class);
+        exception.expectMessage("Input data doesn't correspond require format");
+        parseService.parse(list);
+    }
+
+    @Test
+    public void parse_throwsRuntimeExceptionIfFetchedOperationWasIncorrect_notOk() {
```
zim10093, line 50: the same

src/test/java/service/impl/OperationHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+        fruitDao = new FruitDaoImpl(storage);
+    }
+
+    @Test
+    public void doTransaction_supplyOperationsWorkRight_ok() {
+        String message = "expected quantity and actual are different";
+        OperationHandler operationHandler = new AddOperationHandler(fruitDao);
+        operationHandler.doTransaction(new FruitTransaction("s", "watermelon", 10));
+        assertEquals(message, 10, fruitDao.get("watermelon").intValue());
+    }
```
zim10093, line 39: Is better to chek separatly, in separatet test file , for their works with db.
strategy.AddOperationHandler;
strategy.SetBalanceOperationHandler;
strategy.SubtractOperationHandler;

In this test, you test operation handler and returned operation. so you should test only one thing.
You have method doTransaction. invoke it in operation type class

zim10093 (changes requested):

src/test/java/service/impl/ShopServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+        final String message2 = "failed output data.\n";
+        fruitDao.add("apples", 100);
+        fruitDao.add("oranges", 100);
+        fruitDao.add("nuts", 100);
+        List<String[]> reportStrings = shopService.doReport();
+        assertArrayEquals(message1, reportStrings.get(0), new String[]{"fruit", "quantity"});
+        assertArrayEquals(message2, reportStrings.get(1), new String[]{"oranges", "100"});
+        assertArrayEquals(message2, reportStrings.get(2), new String[]{"nuts", "100"});
+        assertArrayEquals(message2, reportStrings.get(3), new String[]{"apples", "100"});
+        Storage.fruits.clear();
```
zim10093, line 44:
```suggestion
        storage.clear();

```

zim10093 (approved):

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 👍

PR #318, opened 2022-07-26
Homework fruit shop tests

Mykola-Kutsyi:


src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,43 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.Fruit;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class BalanceOperationHandlerTest {
+    private static OperationHandler operationHandler;
+
+    @Before
```
MaliukDaria, line 15: do we need to do this before each test? (same in other places)

src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,43 @@
...
+    public void setUp() {
+        operationHandler = new BalanceOperationHandler(new StorageDaoImpl());
+    }
+
+    @Test
+    public void balance_correct_Ok() {
```
MaliukDaria, line 21:
```suggestion
    public void doOperation_balance_Ok() {
```

There are a lot of ways to name your test methods. The main point is that they should have informative names and be consistent along with other developers in your team. For this task use such convention: `<methodUnderTest>_<state>_<expectedBehavior>`; For example, if we are testing the method register with a `null` user's age the test method name should be `register_nullAge_notOk`. `notOk` is because the test expects the register method to throw an exception.

fix all tests names

src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.FruitTransactionParcer;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ParseServiceImplTest {
```
MaliukDaria, line 13: I think we could add test for several rows with FruitTransaction data

MaliukDaria (changes requested):

PR #334, opened 2022-08-20
Rebuilt the project, added tests to jv-fruit-shop solution

vasiliy-sheretko:


MaliukDaria (changes requested):
Why do you have 3 branches with similar names? I think it is not quite convenient
image
Подебаж тест `supplyWithSuchFruitInStorage_Ok` і подивись чи порожній в тебе storage перед запуском
image

src/test/java/core/basesyntax/DailyReportWriterImplTest.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+import static org.junit.Assert.assertTrue;
+
+import core.basesyntax.service.DailyReportWriter;
+import core.basesyntax.service.impl.DailyReportWriterImpl;
+import java.io.File;
+import org.junit.Test;
+
+public class DailyReportWriterImplTest {
```
aaananas, line 10: add @After method, and clear storage after each test, because i can see that some tests dont clear after

aaananas (changes requested):

kozhukhovsky (changes requested):
Previous comment not fixed.
You added clear in each tests in handler, but it better, if u use method with annotation @after for clearing storage

src/test/java/core/basesyntax/CsvFileReaderTest.java
```diff
@@ -0,0 +1,64 @@
...
+import core.basesyntax.service.Reader;
+import core.basesyntax.service.impl.CsvFileReader;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.Test;
+
+public class CsvFileReaderTest {
+    private final String fromFilePath = "src/test/resources/test-database.csv";
+    private final Reader fileReader = new CsvFileReader();
```
aaananas, line 17:
```suggestion
    private final Reader fileReader;
```
move to a method with annotation Before

aaananas (approved):

PR #437, opened 2022-11-04
Wrote tests for services

OleksandrKorniychuk:


Veronika-Tretiak (approved):

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.HashMap;
+import java.util.Map;
+import myfirstproject.model.Fruit;
+import myfirstproject.service.PreparingData;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class PrepareDataImplTest {
```
hronograf, line 12: let's also check a case when we call `preparingData.prepare(new StringBuilder(), Collections.emptyMap())`
also, why we pass `StringBuilder` as a parameter to `prepare` method?
can we create instance inside `prepare`?

src/test/java/myfirstproject/service/impl/ReadFileImplTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        expectedList.add("type,fruit,quantity".split(","));
+        expectedList.add("b,banana,20".split(","));
+        expectedList.add("b,apple,100".split(","));
+        expectedList.add("s,banana,100".split(","));
+        expectedList.add("p,banana,13".split(","));
+        expectedList.add("r,apple,10".split(","));
+        expectedList.add("p,apple,20".split(","));
+        expectedList.add("p,banana,5".split(","));
+        expectedList.add("s,banana,50".split(","));
```
hronograf, line 70: can we write something like 
```
String expected = "type,fruit,quantity" + System.lineSeparator()
                  + "b,banana,20" + System.lineSeparator()
                  ...;
```

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Map<Fruit, Integer> mapToWrite = new HashMap<>();
+        Fruit fruit = new Fruit("banana");
+        Integer value = 10;
+        mapToWrite.put(fruit, value);
+        writeFile.writeToFile(PATH_TO_NEW_FILE, mapToWrite);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void readWrongFile_not_Ok() {
+        writer.writeToFile(WRONG_PATH, new HashMap<>());
```
hronograf, line 57:
```suggestion
        writer.writeToFile(WRONG_PATH, Collections.emptyMap());
```

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Test
+    public void isCreatedFileExist_Ok() {
+        assertTrue(new File(PATH_TO_NEW_FILE).exists());
+    }
```
hronograf, line 63: `BeforeClass` should setup environment for tests,
we shouldn't test what happened inside `before` method, so let's move that logic to this test

hronograf (changes requested):

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test(expected = RuntimeException.class)
+    public void readWrongFile_not_Ok() {
+        writer.writeToFile(WRONG_PATH, Collections.EMPTY_MAP);
+    }
+
+    @Test
+    public void isCreatedFileExist_Ok() {
+        assertTrue(file.exists());
+    }
```
hronograf, line 65: previous comment wasn't applied
we shouldn't test what happened inside `before` method
image

hronograf (changes requested):

hronograf (changes requested):
also check [this message](https://mateacademy.slack.com/archives/C040S2B99FH/p1667833485897189)

src/test/java/myfirstproject/service/impl/InitializingTest.java
```diff
@@ -0,0 +1,74 @@
...
+import myfirstproject.strategy.PurchaseOperation;
+import myfirstproject.strategy.ReturnOperation;
+import myfirstproject.strategy.SeparationOfOperations;
+import myfirstproject.strategy.SupplyOperation;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.jupiter.api.AfterAll;
+
+public class InitializingTest {
```
hronograf, line 23: let's rename this test
what are we testing here?

src/test/java/myfirstproject/service/impl/InitializingTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void testInitializedOperations_ok() {
+        operation.put(Operation.BALANCE.getOperation(), new BalanceOperation());
+        operation.put(Operation.SUPPLY.getOperation(), new SupplyOperation());
+        operation.put(Operation.PURCHASE.getOperation(), new PurchaseOperation());
+        operation.put(Operation.RETURN.getOperation(), new ReturnOperation());
+        reader.readFile(temporalList, PATH_TO_DB);
+        separation.toDoEachOperation(fruitDao, temporalList, operation);
+        writer.writeToFile(PATH_TO_RESULT, fruitDao.getAll());
+        Assert.assertNotNull(operation);
+    }
```
hronograf, line 48: image

src/test/java/myfirstproject/service/impl/InitializingTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void testBalanceIsNull_ok() {
+        Assert.assertNull(operation.get(Operation.BALANCE.getOperation()));
+    }
```
hronograf, line 53: are we testing `HashMap` here?
we should test own logic, not libraries

src/test/java/myfirstproject/service/impl/ReadFileImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+        fileReaderTest = new ReadFileImpl();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void readWrongFile_notOk() {
+        fileReaderTest.readFile(temporalList, FILE_NOT_EXIST);
+    }
+
+    @Test
+    public void readFile_Ok() {
```
hronograf, line 29:
```suggestion
    public void readFile_fileExists_ok() {
```
check naming everywhere

hronograf (changes requested):

hronograf (changes requested):
in the pom.xml find `excludes` section and change `Main` to `Application`
or rename your `Application.java` file to `Main.java`
if it will not help, tag me in slack private channel
and why didn't you ask mentors to help?

OleksandrKorniychuk:
> in the pom.xml find `excludes` section and change `Main` to `Application` or rename your `Application.java` file to `Main.java` if it will not help, tag me in slack private channel and why didn't you ask mentors to help?

i thought that there was no sense to ask, because it was the last day to pass the project

pom.xml
```diff
@@ -26,6 +26,12 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
+        </dependency>
```
hronograf, line 34: do we need these lines?
let's remove them

OleksandrKorniychuk: we need. 
tests failed after removed.

OleksandrKorniychuk: because 
image

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -0,0 +1,33 @@
+package myfirstproject.service.impl;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
```
hronograf, line 4: here should be 
`import static org.junit.Assert.assertEquals;`
check everywhere (just reimport methods)

hronograf (changes requested):

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
OleksandrKorniychuk: but if we do not have this dependency in pom.xml
then we can't make import . just look:

image

OleksandrKorniychuk: okay... I did it.

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Test
+    public void testInitializedOperations_ok() {
+        operation.put(Operation.BALANCE.getOperation(), BALANCE_OPERATION);
+        operation.put(Operation.SUPPLY.getOperation(), SUPPLY_OPERATION);
+        operation.put(Operation.PURCHASE.getOperation(), PURCHASE_OPERATION);
+        operation.put(Operation.RETURN.getOperation(), RETURN_OPERATION);
+        SEPARATION.toDoEachOperation(FRUIT_DAO, TEMPORAL_LIST, operation);
+    }
```
hronograf, line 49: what are we testing here?

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    public void testBalanceIsNull_ok() {
+        expected = new HashMap<>();
+        expected.put(fruit, value);
+        CustomDataBase.storage.put(fruit, value);
```
hronograf, line 55: do we need this line?
`CustomDataBase.storage.put(fruit, value);`

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -0,0 +1,134 @@
...
+        BALANCE_OPERATION.changeValue(FRUIT_DAO, fruit, value);
+        Integer actual = CustomDataBase.storage.get(fruit);
+        Assert.assertEquals(expected.get(fruit), actual);
+    }
```
hronograf, line 59: it would be better to have an assertion like 
`Assert.assertEquals(expected, CustomDataBase.storage);`
check in all places

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -0,0 +1,134 @@
...
+        Map<Fruit, Integer> expected = FRUIT_DAO.getAll();
+        actual.put(new Fruit("apple"), 10);
+        actual.put(new Fruit("banana"), 10);
+        Assert.assertEquals(expected, actual);
+    }
+
+    @AfterClass
+    public static void after() {
+        operation.clear();
+    }
```
hronograf, line 133: it would be better to clear storage after each test

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -0,0 +1,134 @@
...
+        expected = new HashMap<>();
+        expected.put(fruit, value);
+        CustomDataBase.storage.put(fruit, value);
+        BALANCE_OPERATION.changeValue(FRUIT_DAO, fruit, value);
+        Integer actual = CustomDataBase.storage.get(fruit);
+        Assert.assertEquals(expected.get(fruit), actual);
+    }
+
+    @Test
+    public void testSupplyIsNull_ok() {
```
hronograf, line 62: why do we have `IsNull` in the name?

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+import myfirstproject.model.Fruit;
+import myfirstproject.service.PreparingData;
+import org.junit.Test;
+
+public class PrepareDataImplTest {
+    @Test
+    public void isEmptyData_Ok() {
+        PreparingData preparingData = new PrepareDataImpl();
+        String expected = preparingData.prepare(Collections.emptyMap());
+        String actual = "fruit,quantity" + System.lineSeparator();
```
hronograf, line 17: expected - value that should be
actual - value that method returned 
let's change names

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import myfirstproject.model.Fruit;
+import myfirstproject.service.PreparingData;
+import org.junit.Test;
+
+public class PrepareDataImplTest {
+    @Test
+    public void isEmptyData_Ok() {
+        PreparingData preparingData = new PrepareDataImpl();
```
hronograf, line 15: let's create field with type `PreparingData` and initialise it inside method annotated with `BeforeClass`
also, what about the name of method?
check [this message](https://mateacademy.slack.com/archives/C040S2B99FH/p1667833485897189)

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -0,0 +1,30 @@
...
+        WRITE_FILE.writeToFile(WRONG_PATH, Collections.EMPTY_MAP);
+    }
+
+    @Test
+    public void isCreatedFileExist_Ok() {
+        final File file = new File(PATH_TO_NEW_FILE);
+        Map<Fruit,Integer> testMap = new HashMap<>();
+        WRITE_FILE.writeToFile(PATH_TO_NEW_FILE, testMap);
+        assertTrue(file.exists());
+    }
```
hronograf, line 29: isn't enough for testing
let's also check content of the file (so map shouldn't be empty)

src/main/java/myfirstproject/service/impl/WriteFileImpl.java
```diff
@@ -0,0 +1,21 @@
...
+public class WriteFileImpl implements WriteFile {
+    @Override
+    public void writeToFile(String path, Map<Fruit, Integer> mapToWrite) {
+        PreparingData preparingData = new PrepareDataImpl();
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(path))) {
+            bufferedWriter.write(preparingData.prepare(mapToWrite));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file. " + path, e);
+        }
+    }
```
hronograf, line 20: why are we creating instance of `PreparingData` here?
can we have something like this
```suggestion
    public void writeToFile(String path, String content) {
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(path))) {
            bufferedWriter.write(content);
        } catch (IOException e) {
            throw new RuntimeException("Can't write to file. " + path, e);
        }
    }
```

hronograf (changes requested):

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -46,53 +47,58 @@ public void testInitializedOperations_ok() {
...
+    public void testSupply_ok() {
         expected = new HashMap<>();
-        expected.put(fruit, value + value);
-        CustomDataBase.storage.put(fruit, value);
-        SUPPLY_OPERATION.changeValue(FRUIT_DAO, fruit, value);
-        Integer actual = CustomDataBase.storage.get(fruit);
-        Assert.assertEquals(expected.get(fruit), actual);
+        expected.put(fruitApple, value + value);
+        expected.put(fruitBanana, value);
+        CustomDataBase.storage.put(fruitApple, value);
```
hronograf, line 70: we shouldn't put fruits to our storage directly, because we do it in `SUPPLY_OPERATION.changeValue` 
(check for all methods in this class)

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -46,53 +47,58 @@ public void testInitializedOperations_ok() {
...
+        SUPPLY_OPERATION.changeValue(FRUIT_DAO, fruitApple, value);
+        System.out.println(expected);
+        System.out.println(CustomDataBase.storage);
```
hronograf, line 73: we don't need `System.out.println` in tests

src/test/java/myfirstproject/service/impl/OperationsTests.java
```diff
@@ -114,21 +120,25 @@ public void testDao() {
...
+        actual.put(fruitBanana, value);
         Assert.assertEquals(expected, actual);
     }
 
-    @AfterClass
-    public static void after() {
+    @After
+    public void after() {
         operation.clear();
     }
```
hronograf, line 143: `CustomDataBase.storage.clear()`

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -7,20 +7,28 @@
...
+        preparingData = new PrepareDataImpl();
+    }
+
     @Test
-    public void isEmptyData_Ok() {
-        PreparingData preparingData = new PrepareDataImpl();
-        String expected = preparingData.prepare(Collections.emptyMap());
-        String actual = "fruit,quantity" + System.lineSeparator();
+    public void isEmptyData_ok() {
+        preparingData = new PrepareDataImpl();
```
hronograf, line 23: remove `preparingData = new PrepareDataImpl();`

src/test/java/myfirstproject/service/impl/PrepareDataImplTest.java
```diff
@@ -7,20 +7,28 @@
...
+        String expected = "fruit,quantity" + System.lineSeparator();
+        String actual = preparingData.prepare(Collections.emptyMap());
         assertEquals(expected, actual);
     }
 
     @Test
-    public void isValidPreparedData_Ok() {
-        PreparingData preparingData = new PrepareDataImpl();
+    public void isValidPreparedData_ok() {
+        preparingData = new PrepareDataImpl();
```
hronograf, line 31: same

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -1,30 +1,31 @@
...
     }
 
     @Test
-    public void isCreatedFileExist_Ok() {
-        final File file = new File(PATH_TO_NEW_FILE);
-        Map<Fruit,Integer> testMap = new HashMap<>();
-        WRITE_FILE.writeToFile(PATH_TO_NEW_FILE, testMap);
-        assertTrue(file.exists());
+    public void testContentOfFile_Ok() {
+        ReadFile reader = new ReadFileImpl();
```
hronograf, line 24: image

hronograf (changes requested):

src/test/java/myfirstproject/service/impl/OperationsTests.java
OleksandrKorniychuk: No. We should first put in storage, because he is empty. in this case we will throw NullPointerException. 
If we want to change(changeValue) some value from storage, in first step we should already have some value in this. 
Why out storage in each test empty? because we have @After, where we clear out storage.

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+        assertTrue(file.exists());
+    }
+
+    @Test
+    public void testContentOfFile_ok() {
+        List<String[]> temporalList = new ArrayList<>();
+        WRITE_FILE.writeToFile(PATH_TO_NEW_FILE, Arrays.toString(TEST_CONTENT));
+        temporalList.add(TEST_CONTENT);
+        assertEquals(Arrays.toString(temporalList.get(0)), Arrays.toString(TEST_CONTENT));
+    }
```
hronograf, line 32: what are we testing here?
we don't check what we wrote to file

hronograf (changes requested):

src/test/java/myfirstproject/service/impl/WriteFileImplTest.java
OleksandrKorniychuk: idk, without this test, the code does not pass on the github.
then we have too small coverage

hronograf: we shouldn't write code just to write code
we should understand what is going on
so what are we asserting in this test?

OleksandrKorniychuk: this components using in ReadFile. but if we can't use ReadFile, then I use components from there .

hronograf: you can use `Files.readLines`

OleksandrKorniychuk: But it also read from file.

hronograf: yes, but this way we don't use our services

OleksandrKorniychuk: Aaaaaaa, understood... 😅

hronograf (approved):
Good 🍉

PR #458, opened 2022-11-08
Homework jv fruit shop tests solution

romangolovatyi:


hronograf (commented):
add tests for `ReportGeneratorService` to pass checks

PR #468, opened 2022-12-13
first try on task jv-fruit-shop-tests

romanovosad87:


src/test/java/core/basesyntax/service/impl/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void generateReport_ok() {
+        String actualReport = reportGenerator.generateReport(transactionResultMap);
+        assertEquals(expectedReport, actualReport);
+    }
+
+    @Test
+    public void keyAndValueNull_notOk() {
```
aleksandr-bevza, line 44: How do we check that it is not ok? Do we expect some exception?

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #483, opened 2023-01-10
was solved

Andrii-Varlanovych:


aleksandr-bevza (changes requested):
You should copy your fruit-shop project here and then cover it with tests
image

PR #524, opened 2023-02-04
commit

olei23:


src/test/java/core/basesyntax/service/impl/TransactionParserImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void parseTransactionList_Ok() {
+        List<String> stringList = new ArrayList<>();
+        stringList.add("b,banana,20");
+        stringList.add("b,apple,100");
+        stringList.add("s,banana,100");
+        stringList.add("p,banana,13");
+        stringList.add("r,apple,10");
+        stringList.add("p,apple,20");
+        stringList.add("p,banana,5");
+        stringList.add("s,banana,50");
```
karma-o, line 30: can we move it to some constant var like `TEST_DATA` and initialize it in the `before each`?
we can also use `List.of()` method instead, same with the `expectedList` var

src/test/java/core/basesyntax/service/impl/TransactionParserImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+        expectedList.get(3).setOperation(Transaction.Operation.PURCHASE);
+        expectedList.add(new Transaction("apple", 10));
+        expectedList.get(4).setOperation(Transaction.Operation.RETURN);
+        expectedList.add(new Transaction("apple", 20));
+        expectedList.get(5).setOperation(Transaction.Operation.PURCHASE);
+        expectedList.add(new Transaction("banana", 5));
+        expectedList.get(6).setOperation(Transaction.Operation.PURCHASE);
+        expectedList.add(new Transaction("banana", 50));
+        expectedList.get(7).setOperation(Transaction.Operation.SUPPLY);
+        assertEquals(expectedList.get(2).getOperation(), actual.get(2).getOperation());
```
karma-o, line 49: why assert only 1 element? we can use for loop here

karma-o (changes requested):

src/test/java/core/basesyntax/service/impl/TransactionParserImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+        expectedList.get(6).setOperation(Transaction.Operation.PURCHASE);
+        expectedList.get(7).setOperation(Transaction.Operation.SUPPLY);
+        for (int i = 0; i < 8; i++) {
+            assertEquals(expectedList.get(i).getOperation(), actual.get(i).getOperation());
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void parse_InvalidStoreOperationTransaction_NotOk() {
+        TEST_DATA.add("type,fruit,quantity");
```
andriy-shulepa, line 45: it's better to create a new list based on TEST_DATA and add invalid data to it. because TEST_DATA now can't be reused in new tests since it invalid

src/test/java/core/basesyntax/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @BeforeClass
+    public static void beforeClass() {
+        operationHandler = new BalanceOperationHandler();
+    }
+
+    @Test
+    public void calculate_Ok() {
+        operationHandler.calculate(new Transaction("banana", 20));
+        Integer expectedResult = 20;
```
andriy-shulepa, line 23: generating expected data should be before calling tested method (calculate in this case)

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+
+public class Storage {
+    public static final HashMap<String, Integer> fruitsStorage = new HashMap<>();
```
aaananas, line 6:
```suggestion
    public static final Map<String, Integer> fruitsStorage = new HashMap<>();
```

aaananas (approved):

PR #532, opened 2023-02-14
add solution

uglovyi:


aaananas (changes requested):
there is definitely not enough tests - you should cover all the services with tests

PR #542, opened 2023-02-24
Hw jv fruit shop tests

Iskamele:
During the implementation of the task, I had to edit some parts of the code from the original task, and I also tried Mockito in practice.

Before these decisions and consult with mentors, but did not consult about the details during implementation, so I will be glad if you tell me what I can improve in this implementation without removing Mockito and refactoring all the code without it... 😁

Also, along the work, I heard that using arrange-act-assert comments, and extracting expected-actual into separate variables is a best practice, so I implemented this practice in my test cases.


aaananas (approved):
no need to add those comments
but great job

PR #583, opened 2023-03-17
made tests for services and operations

pechivo:


src/test/java/core/basesyntax/service/impl/FileWriterServiceImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+    public void writeToFile_validFata_Ok() {
+        File file = new File(VALID_FILE_PATH);
+        fileWriterService.writeToFile(REPORT, VALID_FILE_PATH);
+        try {
+            String expected = Files.readAllLines(file.toPath()).get(0);
+            assertEquals(REPORT, expected);
+            file.delete();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
EliSoll, line 39: we don't need to catch exception in tests, you can just throw exception in method's signature))

src/test/java/core/basesyntax/service/impl/FileWriterServiceImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.Files;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class FileWriterServiceImplTest {
+    private static FileWriterService fileWriterService;
+    private static final String HEAD = "fruit,quantity";
+    private static final String INVALID_FILE_PATH = "G://unreal_folder.csv";
+    private static final String VALID_FILE_PATH = "src/test/resources/new.csv";
+    private static String REPORT = HEAD;
```
EliSoll, line 17: Can you use only one constant instead of two here?

EliSoll (approved):
I have left some comments about exception's catching in test's method and constants, but overall - keep going and wish you unproblematic Friday)

src/test/java/core/basesyntax/service/impl/FileReaderServiceImplTest.java
```diff
@@ -0,0 +1,60 @@
...
+        actual = fileReaderService.readFromFile(null);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void readFromFile_inputFileNotExist_notOk() {
+        actual = fileReaderService.readFromFile(NOT_EXISTING_FILE);
+    }
+
+    @Test
+    public void readFromFile_inputFileExist_Ok() {
```
sarakhmen, line 50:
```suggestion
    public void readFromFile_inputFileExist_ok() {
```

src/test/java/core/basesyntax/service/impl/FruitShopServiceImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+        Storage.getFruits().clear();
+    }
+
+    @After
+    public void cleanStorage() {
+        Storage.getFruits().clear();
+    }
+
+    @Test
+    public void processTransaction_withValidInformation_Ok() {
```
sarakhmen, line 42: this class should be responsible only for FruitShopServiceImpl testing. So, let's avoid isolated testing of handlers here. I'd recommend adding instead two additional test cases: when list of transaction == null, and when list is empty

sarakhmen (approved):
Overall good job, you can improve your solution if you want :)

PR #612, opened 2023-04-16
jv-fruitShop-test

SerhiUA:
Added tests;

### DIFF
--- a/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
@@ -1,0 +1,37 @@
+package core.basesyntax.service.impl;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class WriterServiceImplTest {
+    private static final String OUTPUT_FILE_PATH
+            = "src/main/java/core/basesyntax/resources/outputFile.csv";
+    private static String date;
+    private WriterServiceImpl writerService;
+
+    @BeforeEach
+    void setUp() {
+        writerService = new WriterServiceImpl();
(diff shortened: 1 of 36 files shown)


SerhiUA:
Failed to create tests for TransactionServiceImpl. When comparing two identical FruitTransaction issues a fouls. Please tell me, the problem is that I need to override the equals method in FruitTransaction and Fruit

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class WriterServiceImplTest {
+    private static final String OUTPUT_FILE_PATH
+            = "src/main/java/core/basesyntax/resources/outputFile.csv";
+    private static String date;
```
olekskov, line 13:
```suggestion
    private static String data;
```

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+    private WriterServiceImpl writerService;
+
+    @BeforeEach
+    void setUp() {
+        writerService = new WriterServiceImpl();
```
olekskov, line 18: Please do not create new instance of service for each test unless it's really necessary.
On bigger project your test suit may run for several minutes and you won't wish it to run even longer due to creating of a bunch of new objects :)

olekskov (changes requested):

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
olekskov: Still in BeforeEach section, you have it initialized on line 16, that's enough

olekskov (changes requested):

PR #623, opened 2023-05-10
home-work/fruit-shop-tests

andrii-subash:


okuzan (changes requested):
please make sure your build passes before requesting a review 😉

okuzan (changes requested):
please make sure your build passes before requesting a review 😉

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+            assertEquals(expected, actual);
+        } catch (IOException e) {
+            fail();
+        }
+    }
+
+    @Test
+    void createReportFile_WrongPath_NotOk() {
+        assertThrows(IOException.class, () -> Files.readAllLines(
+                Path.of("src/test/resources/data0.txt")));
```
okuzan, line 44: You test library methods, not your own class

src/test/java/core/basesyntax/service/impl/FruitTransactionServiceImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+    void getFruitTransactionsFromFile_Ok() {
+        final List<FruitTransaction> actualList = fruitTransactionService
+                .getFruitTransactionsFromFile("src/test/resources/wrongData.txt");
+
+        final List<FruitTransaction> expectedList = new ArrayList<>();
+
+        FruitTransaction fruitTransaction1 = new FruitTransaction();
+        fruitTransaction1.setOperation(FruitTransaction.Operation.BALANCE);
+        fruitTransaction1.setFruit("banana");
+        fruitTransaction1.setQuantity(20);
```
okuzan, line 27: Your code would be much shorter if you had a constructor for fruitTransation

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+            final List<String> actual = Files.readAllLines(Path.of("src/test/resources/data3.txt"));
+            List<String> expected = new ArrayList<>();
+            expected.add("fruit,quantity");
+            expected.add("banana,100");
+            expected.add("apple,15");
+            assertEquals(expected, actual);
+        } catch (IOException e) {
+            fail();
+        }
+    }
```
okuzan, line 39: You don't test your class' methods here

okuzan (changes requested):

src/test/java/core/basesyntax/service/impl/FruitTransactionServiceImplTest.java
andrii-subash: I agree, but if I create a constructor of FruitTransaction class, then in the FruitTransactionServiceImpl class, the getFruitTransactionsFromFile() method will not be readable when I created the objects of FruitTransaction class, so I use setters there for better readability

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
andrii-subash: image
I disagree because if I change the data the test will fail

PR #628, opened 2023-05-18
all tests implemented

TymurProkhorov:


src/test/java/core/basesyntax/daotest/ReadFromFileTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public static final String FILE_PATH = "src/test/resources/inputOk.csv";
+    public static final String EMPTY_FILE_PATH = "src/test/resources/inputEmpty.csv";
+    public static final String WRONG_FILE_PATH = "src/test/resources/blablabla";
+    public static final String NULL_FILE_PATH = null;
+
+    @Test
+    void readFromFile_Ok() {
+        List<String> expected = new ArrayList<>();
+        fillTransactionsList(expected);
+        ReadFromFile reader = new ReadFromFileImpl();
```
olekskov, line 23: It's not required to initialize service you test for each test if it doesn't hold any state.
Please recheck all test-classes and move used services as class variables where possible

olekskov (changes requested):

src/test/java/core/basesyntax/daotest/ReadFromFileTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final String FILE_PATH = "src/test/resources/inputOk.csv";
+    public static final String EMPTY_FILE_PATH = "src/test/resources/inputEmpty.csv";
+    public static final String WRONG_FILE_PATH = "src/test/resources/blablabla";
+    public static final String NULL_FILE_PATH = null;
+    private ReadFromFile reader;
+
+    @Test
+    void readFromFile_Ok() {
+        List<String> expected = new ArrayList<>();
+        fillTransactionsList(expected);
```
okuzan, line 23: You basically need to compare against a constant, so instead of using those two lines in your tests, you can compare content against a field (created using `List.of` for example)

src/test/java/core/basesyntax/daotest/ReadFromFileTest.java
```diff
@@ -0,0 +1,68 @@
...
+        reader = new ReadFromFileImpl();
```
okuzan, line 24: Fix previous mentor's comment, you need to initialized test service only once, you can use @BeforeClass etc

src/test/java/core/basesyntax/daotest/WriteToFileImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+        writer.writeToFile(report(), FILE_PATH_OK);
+
+        File file = new File(FILE_PATH_OK);
+        String actual;
+        try {
+            actual = String.join(System.lineSeparator(),
+                    Files.readAllLines(file.toPath()));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write to file. " + e);
+        }
```
okuzan, line 32: In test cases, it is generally recommended to let any exceptions propagate to the testing framework, which will handle the exceptions and report them appropriately. By removing the try-catch block, the testing framework can accurately capture and report any exceptions that occur during the test execution.

Therefore, I recommend removing the try-catch block from the test method and letting any IOException propagate naturally. The testing framework will handle and report the exception, providing more accurate information about the failure.

If necessary, you can use assertion methods specific to file handling or exception handling to verify the expected behavior in case of an exception.

src/test/java/core/basesyntax/servicetest/ParserImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+        list.add(fruitTransaction6);
+        list.add(fruitTransaction7);
+        list.add(fruitTransaction8);
+    }
+
+    private List<String> parseList() {
+        return List.of("type,fruit,quantity", "b,banana,50", "b,apple,100",
+                "b,orange,100", "s,banana,150", "s,apple,1000",
+                "s,orange,2200", "p,banana,13", "r,orange,2");
+    }
```
okuzan, line 82: make it a field, method has no logic

okuzan (changes requested):

src/test/java/core/basesyntax/daotest/WriteToFileImplTest.java
okuzan: not fixed

okuzan (changes requested):

okuzan (approved):
Good job!

PR #636, opened 2023-05-26
- Read and Write Service's methods now accept filepath as an argument…

Tupadzaky:
…. This way their implementations don't have to be tied to the path that is passed via constructor;
- Added equals and hashCode to Record model for correct comparing;
- processRecords in FruitServiceImpl ensures that Storage.fruitMap is empty before executing the logic
- parseFileData in CsvReadParserImpl skips the first row of the csv file instead of deleting it;
- Test for all the services implemented


src/test/java/core/basesyntax/service/parser/ReadParserTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    public void parseFileData_emptyList_ok() {
+        List<Record> records = readParser.parseFileData(new ArrayList<>());
+        assertEquals(records, Collections.emptyList());
+    }
+
+    @Test
+    public void parseFileData_null_notOk() {
+        assertThrows(NullPointerException.class, () -> readParser.parseFileData(null));
```
V-Solianyk, line 63: the same case that occurs several times in the code, a mentor's opinion is needed)

V-Solianyk (approved):
great job, there are a few comments but they need clarification from the mentor.

src/test/java/core/basesyntax/service/fruit/FruitServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class FruitServiceTest {
+    private static FruitService fruitService;
+    private List<Record> positiveRecords;
+    private List<Record> negativeRecords;
+    private List<Record> nullRecords;
```
sarakhmen, line 28: I think these fields are better to be removed and created directly in corresponding methods. Also, consider creating a separate test class for each handler

src/test/java/core/basesyntax/service/parser/ReadParserTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+import java.util.NoSuchElementException;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ReadParserTest {
+    private static ReadParser readParser;
+    private List<String> validData;
+    private List<String> invalidOperationsData;
```
sarakhmen, line 21: the same here

src/test/java/core/basesyntax/service/parser/WriteParserTest.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class WriteParserTest {
+    private static WriteParser writeParser;
+    private static final String TYPE_QUANTITY = "type,quantity" + System.lineSeparator();
+    private Map<String, Integer> validData;
```
sarakhmen, line 17: and here

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #642, opened 2023-05-30
commit

Volodymyr-Mykychak:


overpathz (commented):
Please, check task common mistakes first.

ps. You don't need to include your source code to test directory.
In test folder you have to have only separated test classes for each service

okuzan (commented):
.

olekskov (commented):
Good job!

PR #670, opened 2023-06-05
CreateTests

DV7Art:


src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Test;
+
+class ReportServiceImplTest {
+    private static ReportServiceImpl reportService;
+    private static Storage storage;
+
+    @BeforeEach
+    void setUp() {
+        reportService = new ReportServiceImpl();
+        storage = new Storage();
```
maxline, line 17: This two variables can be created just one time. See previous comments about using `BeforeAll` annotation.
You have static methods put() and get() in the Storage class, why we need to create the new instance of Storage?
Please fix all similar cases in all test classes.

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+        String excepted =
+                "fruit,quantity "
+                        + "banana,152 "
+                        + "apple,120";
+        writerService.writeToFile(PATH_TO_REPORT_FILE_TEST, excepted);
+        String actual = validData(PATH_TO_REPORT_FILE_TEST);
+        Assertions.assertEquals(excepted, actual);
+    }
+
+    private String validData(String path) {
```
maxline, line 40:
```suggestion
    private String readData(String path) {
```

src/test/java/core/basesyntax/strategy/impl/PurchaseImplTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @BeforeEach
+    void setUp() {
+        purchase = new PurchaseImpl();
+        fruitTransaction = new FruitTransaction(FruitTransaction.Operation.PURCHASE, "banana", 15);
+    }
+
+    @Test
+    public void purchaseHandler_NegativeQuantity_notOk() {
+        Storage storage = new Storage();
```
maxline, line 25: Why we create a `new Storage()` in every test case? 
You have static methods put() and get() in the Storage class.
Please fix all similar cases in all test classes.

maxline (changes requested):
Good job! Let's improve the solution

rovein (approved):
Good job!

PR #676, opened 2023-06-12
tests done

Hlib13:


okuzan (changes requested):
Please fix the build before requesting a review

src/test/java/core/basesyntax/service/FileReaderTest.java
```diff
@@ -0,0 +1,65 @@
...
+                + "r,apple,10\n"
+                + "p,apple,20\n"
+                + "p,banana,5\n"
+                + "s,banana,50";
+        try (BufferedWriter bufferedWriter =
+                     new BufferedWriter(new FileWriter(new File(inputFile.toURI())))) {
+            bufferedWriter.write(inputData);
+        } catch (IOException e) {
+            throw new RuntimeException("can't write inputData to file: " + inputFile);
+        }
```
kozhukhovsky, line 42: There is no need to write the file. Create it simply and add it to resources. And check only that the text matches when reading.

Hlib13:
Because final version of fruit shop doesn't work
That's why I make a decision make fruit shop again from zero.

src/test/java/core/basesyntax/service/FileReaderTest.java
Hlib13: if you are talking about the "report.сsv" file, then it is filled when the program starts. and the rest I filled out for testing and checking the performance of the program.

olekskov, line 42:
```suggestion
```

olekskov (approved):
Good job overall, approved. 
But writing file in Reader test is just needless. We have inputDatabase.csv predefined for these tests, there is no sense to overwrite it each time

PR #687, opened 2023-06-20
jv-fruit-shop-tests. pochtalon

pochtalon:
Created test class for FruitTransaction, checked constructor for avoiding checking parameters in each service class test;

Created test classes for each Service implementation class; Created test classes for each Handler implementation class; Created 4 files for testing in resources directory;

### DIFF
--- a/src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
@@ -1,0 +1,36 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.ReportService;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class ReportServiceImplTest {
+    private static final String TITLE = "fruit,quantity";
+    private static final String FIRST_PAIR = "banana,152";
+    private static final String SECOND_PAIR = "apple,90";
+    private static ReportService reportService;
+
+    @BeforeAll
+    static void beforeAll() {
+        reportService = new ReportServiceImpl();
+    }
+
+    @AfterAll
+    static void afterAll() {
+        Storage.getStorage().clear();
+    }
+
+    @Test
+    void reportService_Ok() {
+        assertEquals(TITLE, reportService.createReport());
+        Storage.getStorage().put("banana", 152);
+        String expected1 = TITLE + System.lineSeparator() + FIRST_PAIR;
+        assertEquals(expected1, reportService.createReport());
+        Storage.getStorage().put("apple", 90);
+        String expected2 = expected1 + System.lineSeparator() + SECOND_PAIR;
+        assertEquals(expected2, reportService.createReport());
+    }
+}
--- a/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
@@ -1,0 +1,52 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.WriterService;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class WriterServiceImplTest {
+    private static final String PATH_TO_WRITE = "src/test/resources/report.csv";
+    private static final String WRONG_PATH = "src/test/report.csv";
+    private static final String REPORT = "fruit,quantity\n" +
+            "banana,152\n" +
+            "apple,90";
+    private static WriterService writerService;
+
+    @BeforeAll
+    static void beforeAll() {
+        writerService = new WriterServiceImpl();
+    }
+
+    @Test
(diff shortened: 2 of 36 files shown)


src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    void writeServiceValidPathAndReport_Ok() {
+        writerService.writeToFile(PATH_TO_WRITE, REPORT);
+        List<String> expected = List.of("fruit,quantity", "banana,152", "apple,90");
+        try {
+            assertEquals(expected, Files.readAllLines(Paths.get(PATH_TO_WRITE)));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
overpathz, line 49: Write throws in the signature OR
if you catch it, if we did not find a valid path, it means that test is failed, so you can use `Assertions.fail(e);`

src/test/java/core/basesyntax/strategy/operations/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,36 @@
...
+import core.basesyntax.strategy.Operation;
+import core.basesyntax.strategy.OperationHandler;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class BalanceOperationHandlerTest {
+    private static final Operation VALID_OPERATION = Operation.BALANCE;
+    private static final String VALID_FRUIT = "apple";
+    private static final int VALID_QUANTITY = 20;
```
overpathz, line 16: Consider add tests to check:
what if we put null fruit, zero or negative quantity?

Then, if we will have several tests working with Storage, to check them independently, we have to clear storage in `@AfterEach` method

src/test/java/core/basesyntax/strategy/operations/PurchaseOperationHandlerTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.strategy.Operation;
+import core.basesyntax.strategy.OperationHandler;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class PurchaseOperationHandlerTest {
```
overpathz, line 14: What if we do purchase with (illegal) negative quantity?

src/test/java/core/basesyntax/strategy/operations/ReturnOperationHandlerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.strategy.Operation;
+import core.basesyntax.strategy.OperationHandler;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ReturnOperationHandlerTest {
```
overpathz, line 13: what if we put transaction to not empty storage?

src/test/java/core/basesyntax/strategy/operations/SupplyOperationHandlerTest.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final int SUPPLY_QUANTITY = 10;
+    private static OperationHandler supplyHandler;
+
+    @BeforeAll
+    static void beforeAll() {
+        supplyHandler = new SupplyOperationHandler();
+        Storage.getStorage().put(VALID_FRUIT, VALID_QUANTITY);
+    }
+
+    @AfterAll
```
overpathz, line 26: image

Since you can have multiple tests (or you will expand your logic so the tests too), you have to use this way of storage cleaning

overpathz (changes requested):

src/test/java/core/basesyntax/strategy/operations/BalanceOperationHandlerTest.java
pochtalon: If I already checked "null fruit, zero or negative quantity" in FruitTransaction constructor, I can't to get invalid FruitTransaction object. So should I check these fields in each Handler's tests?

src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+        reportService = new ReportServiceImpl();
+    }
+
+    @AfterAll
+    static void afterAll() {
+        Storage.getStorage().clear();
+    }
+
+    @Test
+    void reportService_Ok() {
```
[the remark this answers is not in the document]
kozhukhovsky: not fixed

src/test/java/core/basesyntax/strategy/operations/BalanceOperationHandlerTest.java
kozhukhovsky: Not necessarily, but you can make sure that there will be an exception here if the value is not valid.

kozhukhovsky (changes requested):
Other comments need to be corrected

src/test/java/core/basesyntax/strategy/operations/BalanceOperationHandlerTest.java
pochtalon: image

I created this checks for each Handler class, but I had made alredy checking  in FruitTransaction constuctor, so I can't create invalid FruitTransaction objects. 
So, should I refuse from invalid cases in handlers, or in FruitTransaction's constructor?

kozhukhovsky (approved):

PR #688, opened 2023-06-20
jv-fruit-shop-part-2

s-hrynus:


src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+        readerService.readFromFile(INPUT_CSV_WITH_CONTENT);
+        File file = new File(INPUT_CSV_WITH_CONTENT);
+        Assertions.assertTrue(file.exists());
+    }
+
+    @Test
+    void reader_returnCorrectData_notOk() {
+        List<String> actual = List.of("c,car,blue", "s,truck,green", "m,plane,320");
+        List<String> expected = readerService.readFromFile(INPUT_CSV_WITH_CONTENT);
+        Assertions.assertNotEquals(expected, actual);
```
okuzan, line 67: it doesn't make sense to create a negative test for correct content. There are endless possibilities of wrong result, you should just test that it's equal to expected

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    void reader_validFilePath_Ok() {
+        readerService.readFromFile(INPUT_CSV_WITH_CONTENT);
+    }
+
+    @Test
+    void reader_fileCreatedSuccessfully_Ok() {
+        readerService.readFromFile(INPUT_CSV_WITH_CONTENT);
+        File file = new File(INPUT_CSV_WITH_CONTENT);
+        Assertions.assertTrue(file.exists());
```
okuzan, line 60: why are you testing that file exists in ReaderService? Even if your test file didn't exist, it would only mean that you have written the test wrong, not that your service doesn't work :)

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    void reader_emptyData_notOk() {
+        Assertions.assertThrows(TransactionException.class,
+                () -> readerService.readFromFile(EMPTY_CSV),
+                String.format("%s should throw when input file is empty", EXCEPTION_MESSAGE));
+    }
+
+    @Test
+    void reader_validFileFormat_notOk() {
```
okuzan, line 44: valid notOk?

src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+    void report_emptyStorage_notOk() {
+        Assertions.assertThrows(TransactionException.class, () -> reportService.createReport(),
+                String.format("%s should throw for empty storage", EXCEPTION_MESSAGE));
+    }
+
+    @Test
+    void report_isNotEmptyStorage_Ok() {
+        Storage.fruits.put("banana", 20);
+        Storage.fruits.put("apple", 10);
+        Assertions.assertNotEquals(HEADER, reportService.createReport());
```
okuzan, line 33: don't create negative test here, the fact that report isn't just header tell us little about it behaving correctly. Maybe the report is actually "Encountered unexpected error on line 47" or whatever

src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    void report_createCorrectReport_Ok() {
```
okuzan, line 37:
```suggestion
    void report_validData_ok() {
```

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    void writer_contentIsNull_notOk() {
+        Assertions.assertThrows(TransactionException.class,
+                () -> writerService.write(null, CORRECT_FILE_PATH),
+                String.format("%s should throw for report content which is null",
+                        EXCEPTION_MESSAGE));
+    }
+
+    @Test
+    void writer_validPath_notOk() {
```
okuzan, line 41: valid notOk

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+                () -> writerService.write(REPORT_CONTENT, INCORRECT_FILE_PATH),
+                String.format("%s should throw for incorrect file path",
+                        EXCEPTION_MESSAGE));
+    }
+
+    @Test
+    void writer_fileCreatedCorrect_Ok() {
+        writerService.write(REPORT_CONTENT, CORRECT_FILE_PATH);
+        File file = new File(CORRECT_FILE_PATH);
+        Assertions.assertTrue(file.exists());
```
okuzan, line 52: If you want to have multiple positive tests, then write the expected result instead of ok
```suggestion
    void writer_validData_fileCreated() {
        writerService.write(REPORT_CONTENT, CORRECT_FILE_PATH);
        File file = new File(CORRECT_FILE_PATH);
        Assertions.assertTrue(file.exists());
```

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Test;
+
+class WriterServiceImplTest {
+    private static final String EXCEPTION_MESSAGE
+            = TransactionException.class.toString();
+    private static final String CORRECT_FILE_PATH = "src/main/resources/output.csv";
+    private static final String INCORRECT_FILE_PATH
+            = "C:/Users/.../my-project/src/main/resources/output.csv";
+    private static final String REPORT_CONTENT = "apple,120";
+    private static WriterService writerService;
```
okuzan, line 17: In general, it is more idiomatic to declare the service being tested as an instance variable rather than a static variable.

src/test/java/core/basesyntax/strategy/operation/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    void handler_correctBalance_Ok() {
+        operationHandler.handle(fruitTransaction);
+        Integer expected = fruitTransaction.getQuantity();
+        Integer actual = Storage.fruits.get("apple");
+        Assertions.assertEquals(expected, actual);
+    }
+
+    @Test
+    void handler_correctBalance_notOk() {
```
okuzan, line 49: correct notOk?

src/test/java/core/basesyntax/strategy/operation/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,55 @@
...
+        operationHandler.handle(fruitTransaction);
+        Integer expected = fruitTransaction.getQuantity();
+        Integer actual = Storage.fruits.get("banana");
+        Assertions.assertNotEquals(expected, actual);
```
okuzan, line 53: it's not really expected if you assert they are not equal right 😄. Don't just write negative tests so there are more tests

src/test/java/core/basesyntax/strategy/operation/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,100 @@
...
+    static void beforeAll() {
+        operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(FruitTransaction.Operation.BALANCE,
+                new BalanceOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.SUPPLY,
+                new SupplyOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.RETURN,
+                new ReturnOperationHandler());
```
okuzan, line 28: use Map.of to do it in less lines

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
s-hrynus: Yeah, totally agree. To my mind when I wrote that test, I think how to do a lot as tests as I can. So, as a result, I added code for creating file in class which should just read from file :(

src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+import core.basesyntax.service.TransactionException;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ParseServiceImplTest {
+    private static final String EXCEPTION_MESSAGE
+            = TransactionException.class.toString();
```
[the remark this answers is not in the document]
s-hrynus: Directly it means I should use in Assertions.assertThrows() the third parameter string or syntaxis should look below?

Assertions.assertThrows(TransactionException.class, () -> parseService.parse(records));

src/test/java/core/basesyntax/strategy/operation/BalanceOperationHandlerTest.java
s-hrynus: My bad, I am trying to get data witch isn't present in storage :(

PR #689, opened 2023-06-23
Solution

OlegMR0:


Ivan95kos (changes requested):
Move strategy and handler testing from TransactionHandlerImplTest into separate classes

aaananas (approved):

PR #692, opened 2023-07-04
Solution of this task

IvanovYurii:


src/test/java/core/basesyntax/service/impl/DataValidatorServiceImplTest.java
```diff
@@ -0,0 +1,152 @@
...
+    private static final List<String> VALID_DATA = Arrays.asList("type,fruit,quantity",
+            "b,banana,20",
+            "s,banana,100",
+            "p,banana,13",
+            "r,apple,10");
+    private final DataValidatorService dataValidatorService =
+            new DataValidatorServiceImpl();
+
+    @Test
+    public void test_DataValidator_DataIsValid_Ok() {
```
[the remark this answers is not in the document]
kozhukhovsky: isDataValid_validData_ok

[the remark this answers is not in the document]
kozhukhovsky: all other tests must also be renamed according to the convention

src/test/java/core/basesyntax/service/impl/DataValidatorServiceImplTest.java
```diff
@@ -0,0 +1,152 @@
...
+        List<String> mutableData = new ArrayList<>(VALID_DATA);
+        assertTrue(dataValidatorService.isDataValid(mutableData));
+    }
+
+    @Test
+    public void test_DataValidator_DataIsNull_NotOk() {
+        List<String> mutableData = null;
```
kozhukhovsky, line 30: you have this variable in each test, move this to the class field and init in each test

kozhukhovsky (changes requested):

src/test/java/core/basesyntax/service/operation/impl/OperationSupplyHandlerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertEquals(30, operationDao.get("banana"));
+        fruitTransaction.setQuantity(5);
+        operationHandler.processOperation(fruitTransaction);
+        assertEquals(35, operationDao.get("banana"));
+    }
+
+    @Test
+    public void test_OperationSupply_NotFruitInDB_NotOk() {
+        assertEquals(0, operationDao.get("apple"));
+    }
```
okuzan, line 41: This test case is unnecessary as it doesn't directly pertain to the OperationSupplyHandler behavior. The responsibility of the handler is to process transactions, not to ensure the initial state of the data.

src/test/java/core/basesyntax/service/operation/impl/OperationSupplyHandlerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    public void test_OperationSupply_FruitInDB_Ok() {
+        assertNotNull(operationDao.get("banana"));
+        assertNotEquals(0, operationDao.get("banana"));
+        assertEquals(20, operationDao.get("banana"));
+    }
```
okuzan, line 48: Similar to the previous point, this assertion seems to be testing the behavior of the OperationDao rather than the OperationSupplyHandler. This could be removed

okuzan (changes requested):
Fix in each handler

Ivan95kos (approved):
Nice

PR #704, opened 2023-07-20
Solution

ArtEg0:
Copied solution from previous task.
Added test classes.

### DIFF
--- a/src/test/java/dao/FruitStorageDaoTests.java
+++ b/src/test/java/dao/FruitStorageDaoTests.java
@@ -1,0 +1,167 @@
+package dao;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.HashMap;
+import java.util.Map;
+import model.Fruit;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class FruitStorageDaoTests {
+    private FruitStorageDao fruitStorageDao;
+
+    @BeforeEach
+    void setUp() {
+        fruitStorageDao = new FruitStorageDaoImpl();
+
+        fruitStorageDao.set("banana", 100);
+        fruitStorageDao.set("apple", 25);
+    }
+
+    @Test
+    void getFruit_ok() {
+        String fruitName = "banana";
+        int fruitQuantity = 100;
+        Fruit expected = new Fruit(fruitName, fruitQuantity);
+        Fruit actual = fruitStorageDao.getFruit("banana");
+
+        assertEquals(expected.getName(), actual.getName(),
+                "Actual and expected fruit name must be equals");
+
+        assertEquals(expected.getQuantity(), actual.getQuantity(),
+                "Actual and expected fruit q-ty must be equals");
+    }
+
+    @Test
+    void getInvalidFruit_notOk() {
+        String fruitName = "coconut";
+        assertThrows(RuntimeException.class, () -> fruitStorageDao.getFruit(fruitName),
+                "There is no such fruit: " + fruitName);
+    }
+
+    @Test
+    void set_ok() {
+        String fruitName = "banana";
+        int fruitQuantity = 50;
+        Fruit expected = new Fruit(fruitName, fruitQuantity);
+        fruitStorageDao.set("banana", 50);
+        Fruit actual = fruitStorageDao.getFruit("banana");
+
+        assertEquals(expected.getName(), actual.getName(),
+                "Actual and expected fruit names must be equals");
+        assertEquals(expected.getQuantity(), actual.getQuantity(),
+                "Actual and expected fruit q-ty must be equals");
+    }
+
+    @Test
+    void setNegativeQuantity_notOk() {
+        fruitStorageDao.set("banana", -100);
+        assertThrows(RuntimeException.class, () -> fruitStorageDao.getFruit("banana"),
+                "Stored quantity can't be less than zero");
+    }
+
+    @Test
+    void add_ok() {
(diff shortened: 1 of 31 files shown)


src/test/java/service/ServiceTests.java
```diff
@@ -0,0 +1,183 @@
...
+
+class ServiceTests {
+    private static final String TEST_FILE = "src/main/resources/test.csv";
+    private ReaderService reader;
+    private WriterService writer;
+    private InputDataResolver resolver;
+    private TransactionOperation transactionOperation;
+    private GenerateReport report;
+    private FruitStorageDao fruitStorageDao;
+    private Path testFilePath;
```
olekskov, line 36: Unit-test class should test only one implementation. And your test classes tree should have absolutely same structure as main package in order to make understanding of what it tested easier.
Please separate and place your tests properly.
e.g.
we have main/java/service/impl/WriterServiceImpl.java
that means that tests for it are expected as test/java/service/impl/WriterServiceImplTest.java

olekskov (changes requested):
Please fix project structure before we may proceed with more in-depth review

src/test/java/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+class WriterServiceImplTest {
+    private static final String TEST_FILE = "src/main/resources/test.csv";
+    private WriterService writer;
+    private FruitStorageDao fruitStorageDao;
+    private ReaderService reader;
+    private Path testFilePath;
+
+    @BeforeEach
+    void setUp() {
```
okuzan, line 26: Don't use try catch throw in unit tests, use assertThrows if you expect exception. Unexpected exception _will_ fail the test, no need to rethrow anything, this isn't business logic (where we catch something specific and wrap it in our customary exception)

src/test/java/model/FruitTest.java
```diff
@@ -0,0 +1,29 @@
...
+    private Fruit fruit;
+
+    @BeforeEach
+    void setUp() {
+        fruit = new Fruit("banana", 125);
+    }
+
+    @Test
+    void getName_ok() {
+        Fruit expected = new Fruit("banana", 125);
```
okuzan, line 18: you don't need to test models, they are tested indirectly when you are testing your services (you use models there)

src/test/java/strategy/BalanceOperationStrategyTest.java
```diff
@@ -0,0 +1,39 @@
...
+    void setUp() {
+        fruitStorageDao = new FruitStorageDaoImpl();
+        balanceOperationStrategy = new BalanceOperationStrategy();
+        fruitStorageDao.set("banana", 15);
+        fruitStorageDao.set("apple", 10);
+    }
+
+    @Test
+    void handleBalanceOperation_ok() {
+        Map<String,Integer> expected = new HashMap<>();
```
okuzan, line 27: what happens if we try to set negative values?

src/test/java/strategy/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+        operationStrategy = new OperationStrategyImpl();
+    }
+
+    @Test
+    void getBalanceOperation_ok() {
+        new BalanceOperationStrategy();
+        OperationHandler handler;
+        handler = operationStrategy.get(Operation.BALANCE);
+
+        assertTrue(handler instanceof BalanceOperationStrategy,
```
okuzan, line 23:
```suggestion
        OperationHandler handler = operationStrategy.get(Operation.BALANCE);
        
        assertTrue(handler instanceof BalanceOperationStrategy,
```
fix everywhere

src/test/java/strategy/PurchaseOperationStrategyTest.java
```diff
@@ -0,0 +1,39 @@
...
+    @BeforeEach
+    void setUp() {
+        fruitStorageDao = new FruitStorageDaoImpl();
+        purchaseOperationStrategy = new PurchaseOperationStrategy();
+        fruitStorageDao.set("banana", 15);
+        fruitStorageDao.set("apple", 10);
+    }
+
+    @Test
+    void handlePurchaseOperation_ok() {
```
okuzan, line 26: what if we try to purchase more than we have? add negative test

src/test/java/dao/FruitStorageDaoTests.java
```diff
@@ -0,0 +1,167 @@
...
+        String fruitName = "banana";
+        int fruitQuantity = 50;
+        Fruit expected = new Fruit(fruitName, fruitQuantity);
+        fruitStorageDao.set("banana", 50);
+        Fruit actual = fruitStorageDao.getFruit("banana");
+
+        assertEquals(expected.getName(), actual.getName(),
+                "Actual and expected fruit names must be equals");
+        assertEquals(expected.getQuantity(), actual.getQuantity(),
+                "Actual and expected fruit q-ty must be equals");
```
okuzan, line 55: generate equals and hashcode for fruit, and compare each fruit field at once

okuzan (changes requested):

src/test/java/strategy/BalanceOperationStrategyTest.java
ArtEg0: it will except the negative value, but after that FruitStorageDaoImpl will throw an exception that q-ty should be greater or equal to zero..

src/test/java/dao/FruitStorageDaoTests.java
ArtEg0: > generate equals and hashcode for fruit, and compare each fruit field at once

Didn't know I can modify source code. Thx)

okuzan (approved):
Good job!

PR #758, opened 2023-08-24
Hw jv fruit shop tests

hanshyn:
Added tests. Check mvn clean verify


okuzan (changes requested):
We don't need to test models like fruitTransaction explicitly, they are used in services, handlers - i.e. test implicitly. Add negative test to handlers, What happens if we submit invalid data? Remember, you can make changes to src/main at this stage if you only found out about something you want to change.

kozhukhovsky (approved):

PR #760, opened 2023-08-24
completed java fruit shop tests task

Ka1zerov:


src/test/java/service/impl/CsvReaderServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+                READER_SERVICE.readFromFile(NO_FILE_NAME));
+    }
+
+    @Test
+    public void emptyFile_Ok() {
+        assertEquals(0,
+                READER_SERVICE.readFromFile(EMPTY_FILE_NAME).length,
+                "Expected array length is 0");
+    }
+
```
okuzan, line 34:
```suggestion
```

src/test/java/service/impl/CsvWriterServiceImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+            throw new RuntimeException("No such file ", e);
+        }
+    }
+
+    @Test
+    public void noFile_notOk() {
+        assertThrows(RuntimeException.class, () ->
+                WRITER_SERVICE.writeToFile(NO_FILE_NAME, REPORT_MESSAGE));
+    }
+
```
okuzan, line 36:
```suggestion
```

src/test/java/service/impl/ReportCreatorImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    public void defaultStorage_Ok() {
+        Map<String, Integer> fruitBox = new HashMap<>();
+        fruitBox.put("banana", 10);
+        fruitBox.put("apple", 10);
+        assertEquals(EXPECTED_DEFAULT_MESSAGE,
+                REPORT_CREATOR.createReport(new Storage(fruitBox)),
+                "Report does not match the expected message");
+    }
+
```
okuzan, line 49:
```suggestion
```

src/test/java/strategy/impl/PurchaseTypeHandlerTest.java
```diff
@@ -0,0 +1,28 @@
+package strategy.impl;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.Test;
+
+public class PurchaseTypeHandlerTest {
+
+    private static final PurchaseTypeHandler PURCHASE_TYPE_HANDLER = new PurchaseTypeHandler();
```
okuzan, line 9: In Java, constant naming conventions (using all uppercase letters with underscores) are typically reserved for immutable values, such as static final primitive values or immutable objects.

The PURCHASE_TYPE_HANDLER instance is not an immutable constant in the truest sense, even though it is marked as final. It refers to an instance of a class, and unless the class itself is designed to be immutable, the object could potentially have mutable state.

A better convention for this case would be to use camelCase for the field name.

src/test/java/db/StorageFillerImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import strategy.ActivityTypeHandler;
+import strategy.ActivityTypeStrategy;
+import strategy.impl.ActivityTypeStrategyImpl;
+import strategy.impl.AdditionTypeHandler;
+import strategy.impl.BalanceTypeHandler;
+import strategy.impl.PurchaseTypeHandler;
+
+public class StorageFillerImplTest {
```
okuzan, line 23: we don't need to test it

src/test/java/service/impl/ActivityParserServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.List;
+import model.Activity;
+import org.junit.Test;
+import service.ActivityParserService;
+
+public class ActivityParserServiceImplTest {
+    private static final ActivityParserService PARSER_SERVICE = new ActivityParserServiceImpl();
+    private static final String[] EMPTY_STRING = {};
+    private static final String[] DEFAULT_STRING = {
```
okuzan, line 15:
```suggestion
    private static final String[] TRANSACTIONS = {
```

src/test/java/service/impl/ActivityParserServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+            "b,apple,100",
+            "s,banana,100",
+            "p,banana,13",
+            "r,apple,10",
+            "p,apple,20",
+            "p,banana,5",
+            "s,banana,50",
+            "b,kiwi,10"
+    };
+    private static List<Activity> activityList;
```
okuzan, line 26: no need to include data type in the variable's name, Java is a strongly typed language

src/test/java/service/impl/ActivityParserServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+        activityList = PARSER_SERVICE.getActivityList(DEFAULT_STRING);
+        assertEquals(9, activityList.size(), "Expected activity list size is 9");
+    }
+
+    @Test
+    public void nullStringArray_NotOk() {
+        assertThrows(NullPointerException.class, () ->
+                PARSER_SERVICE.getActivityList(null));
+    }
+
```
okuzan, line 45:
```suggestion
```

src/test/java/service/impl/CsvReaderServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final String DEFAULT_FILE_NAME = "src/test/java/test_files/defaultFile.csv";
+    private static final String NO_FILE_NAME = "";
+    private static final CsvReaderService READER_SERVICE = new CsvReaderServiceImpl();
+
+    @Test
+    public void defaultFile_Ok() {
+        assertEquals(9,
+                READER_SERVICE.readFromFile(DEFAULT_FILE_NAME).length,
+                "Expected array length is 9");
+    }
```
okuzan, line 20: better assert actual contents

src/test/java/service/impl/CsvWriterServiceImplTest.java
```diff
@@ -0,0 +1,37 @@
...
+    public void defaultFile_Ok() {
+        WRITER_SERVICE.writeToFile(TO_FILE_NAME, REPORT_MESSAGE);
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(TO_FILE_NAME))) {
+            assertEquals(REPORT_MESSAGE,
+                    reader.readLine(),
+                    "File content does not match the expected message");
+        } catch (IOException e) {
+            throw new RuntimeException("No such file ", e);
+        }
```
okuzan, line 28: we don't need to use try catch in any test. If exception occurs unexpectedly - test fails automatically, if we expect it - we can use assertthrows

okuzan (changes requested):

src/test/java/db/StorageFillerImplTest.java
Ka1zerov: why not?

olekskov (approved):

PR #771, opened 2023-09-13
HW Fruit shop tests v1.1

vrudkovsky:


olekskov (changes requested):
Please fix build in order to proceed with review.
Not all tests are passed

PR #772, opened 2023-09-13
implemented test for services and strategies

ystankevych:


src/test/java/core/basesyntax/service/WriterServiceTest.java
```diff
@@ -0,0 +1,39 @@
...
+import core.basesyntax.service.impl.WriterServiceImpl;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class WriterServiceTest {
+    private static final String VALID_FILE_NAME = "src/test/resources/report.csv";
+    private static final String INVALID_FILE_NAME = "src/tests/resources/report.csv";
```
Oleksandr-Tymoshenko, line 15: maybe you forgot to change path?

Oleksandr-Tymoshenko (approved):
Good job, i recommend you also add tests for null values

src/test/java/core/basesyntax/service/WriterServiceTest.java
ystankevych: Hi. No )) Look valid file is in the test directory and the invalid one is in non existing "tests" directory. But I think it would be better to change this path to another one just to make it clearer for another developer =)

ystankevych: done

src/test/java/core/basesyntax/service/CsvParserServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("invalidRowFormatData")
+    void parse_ListWithInvalidRowFormat_NotOk(List<String> invalidData) {
+        assertThrows(RuntimeException.class, () -> csvParserService.parse(invalidData),
+                "Invalid number of columns in a row");
+    }
+
+    private static Stream<List<String>> invalidHeaderData() {
```
Elena-Bruyako, line 90: It's better to replace data for parametrised tests before tests method

Elena-Bruyako (approved):
Good job! Left one minor comment.

PR #775, opened 2023-09-15
Hw fruit shop test

DZarudnyi:


DZarudnyi:
Had some hard time writing tests for void methods, especially WriterImplTest and FruitTransactionTest. I had to use other classes to test these ones, while lesson stated that i should not use other services to test. But it seems like there is no other option to do it?

olekskov (approved):

PR #782, opened 2023-09-18
implemented tests for fruitShop

Parashchuk:


vsych321 (approved):
Great job! There is nothing to add, but as for me there are a lot of tests

src/test/java/core/basesyntax/service/WriterServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+        String actual;
+
+        try {
+            actual = Files.readAllLines(path).stream()
+                    .collect(Collectors.joining(System.lineSeparator()));
+        } catch (IOException e) {
+            throw new RuntimeException(
+                    "Try to write to a file was unsuccessful: " + e
+            );
+        }
```
liliia-ponomarenko, line 45: better move it to private method and add cases with empty file and wrong file path

src/test/java/core/basesyntax/strategy/PurchaseOperationTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import core.basesyntax.db.FruitStorage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Operation;
+import core.basesyntax.strategy.operations.PurchaseOperation;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class PurchaseOperationTest {
```
liliia-ponomarenko, line 13: you also should add a case when fruits are not enough  for purchase

liliia-ponomarenko (changes requested):
good job! Let’s  fix a few mistakes ;)

Elena-Bruyako (approved):

PR #788, opened 2023-09-19
Tests done

Gagadu:


src/test/java/core/basesyntax/service/impl/FileWriterServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    public void writeToFile_correctPath_ok() {
+        fileWriterService.writeToFile(testData, CORRECT_FILE_PATH);
+        List<String> dataFromFile;
+        try {
+            dataFromFile = Files.readAllLines(Path.of(CORRECT_FILE_PATH));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file " + CORRECT_FILE_PATH, e);
+        }
```
okuzan, line 37: no need to use try catch in these tests. If any exception is thrown - the test fails

src/test/java/core/basesyntax/strategy/operation/util/TransactionHandlerTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    public void incomingTransaction_supplyNewFruit_ok() {
+        transactionHandler.processIncomingTransaction(appleSupplyTransaction);
+        Assert.assertEquals(appleSupplyTransaction.getSum(), Storage.storage.get(apple));
+    }
+
+    @Test
+    public void incomingTransaction_supplyExistedFruit_ok() {
```
okuzan, line 52:
```suggestion
    public void incomingTransaction_supplyExistingFruit_ok() {
```
change everywhere

okuzan (changes requested):
Good job overall!

PR #819, opened 2023-10-26
Added solution

slaybrute:


olekskov (changes requested):
please fix build before we may proceed with review

PR #821, opened 2023-10-27
Roma Menzheres's realization of "Java Fruit Store Tests" task

RomanMenzheres:
Created tests with coverage more than 80%


olekskov (approved):
Good job overall, approved
Next times please make packages for tests same as in main code.
So if you have parser in main/service package then tests for it will be in tests/service etc.

PR #822, opened 2023-10-28
jv-fruit-shop-tests HW solution

luk14n:


luk14n:
image
All local checks passed

luk14n:
image
Same problem, all Maven checks works just fine locally

PR #825, opened 2023-10-30
jv-fruit-shop-tests solution

luk14n:


luk14n:
P.S. sorry for multiple prs, was having different results when testing locally and here with GitHub CI. Btw what can be cause of that behaviour?

kerrrusha:
> P.S. sorry for multiple prs, was having different results when testing locally and here with GitHub CI. Btw what can be cause of that behaviour?

can't get definite answers without checking logs

PR #852, opened 2023-11-09
FruitShopTest-hw-solution

IhorTrokhymchuk:


olekskov (approved):
Good job overall, approved.

For future: files required for testing should be in resources folder of test package, not project root

PR #881, opened 2023-12-27
Hw 22 solution

andriimikhaylovskiy:


src/test/java/core/basesyntax/service/impl/CsvParserServiceTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class CsvParserServiceTest {
+    private CsvParserService csvParserService;
+
+    @BeforeEach
+    void setUp() {
+        csvParserService = new CsvParserService();
+    }
```
olekskov, line 19: for all test classes:
try to initialize services required for testing once (during declaration or in BeforeAll method) as when test-suite becomes bigger unnecessary creation of objects affects time to run tests

src/test/java/core/basesyntax/service/impl/CsvReaderServiceTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @BeforeEach
+    void setUp() {
+        csvReaderService = new CsvReaderService();
+    }
+
+    @Test
+    void read_file_ok() {
+        csvReaderService.readFromFile("src/main/resources/input.csv");
+    }
```
olekskov, line 19: we need to check result of reading, it easily can return empty string or any other value that do not correspond file content

src/test/java/core/basesyntax/service/impl/CsvWriterServiceTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @BeforeEach
+    void setUp() {
+        csvWriterService = new CsvWriterService();
+    }
+
+    @Test
+    void write_file_ok() {
+        csvWriterService.writeToCsv("blabla", "src/main/resources/output.csv");
+    }
```
olekskov, line 19: 1. same as above
2. please create and use test resources for testing

olekskov (changes requested):

src/test/java/core/basesyntax/service/impl/CsvWriterServiceTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @BeforeEach
+    void setUp() {
+        csvWriterService = new CsvWriterService();
+        csvReaderService = new CsvReaderService();
+    }
+
+    @Test
+    void write_file_ok() {
+        String content = "blabla";
+        String outputPath = "src/main/resources/output.csv";
```
olekskov, line 25: Next time do not make changes to main package from tests
```suggestion
        String outputPath = "src/test/resources/output.csv";
```

olekskov (approved):
Good job

PR #888, opened 2024-01-15
created test methods

KhrystynaDidykh:


src/test/java/core/basesyntax/model/fruittransaction/FruitTransactionTest.java
```diff
@@ -0,0 +1,52 @@
+package core.basesyntax.model.fruittransaction;
```
olekskov, line 1: There is no need in separate tests for models.
Their getters/setters/constructors are naturally tested when used in other test classes

olekskov (approved):
Good job!

Next time you shouldn't write tests for models

PR #897, opened 2024-01-24
added tests

RCherniavskiy:


RCherniavskiy:
There are no errors in Intelij Idea. What's wrong?

olekskov (changes requested):

src/test/java/core/basesyntax/service/strategy/PurchaseHandlerTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import core.basesyntax.dao.FruitDao;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class PurchaseHandlerTest {
+    @Test
+    void purchaseProcessOperationIs_Ok() {
+        try {
```
olekskov, line 16: Please remove all try/catch constructions from tests.
If exception occures during test perform it's logic then it's automatically failed (unless we have assertThows construction).
Even if we have some checked exception in logic we usually just mention it in test method header

```suggestion
```

src/test/java/core/basesyntax/service/impl/FruitTransactionImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    void processBalanceIs_Ok() {
+        try {
+            List<String> inputLines = Files.readAllLines(Path.of(TEST_FILE_PATH));
+        } catch (Exception e) {
+            fail("Can't get data from file " + TEST_FILE_PATH, e);
+            return;
+        }
```
olekskov, line 52: you never use inputLines variable, why do we need it?

olekskov (changes requested):

olekskov (approved):

PR #908, opened 2024-02-21
First Attempt. Implemented Tests with 91% coverage

kacper-kukula:


src/test/java/core/basesyntax/db/StorageTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    void getFruitStorage_getInstance_Ok() {
+        Map<String, Integer> storage = Storage.getFruitStorage();
+        assertNotNull(storage);
+        assertSame(storage, Storage.getFruitStorage());
+    }
+
+    @Test
+    void getFruitStorage_isHashMap_Ok() {
```
nowickimj, line 27: In my opinion, tests like this one are unnecessary as they are too detailed - please focus on business details rather than technical ones.

src/test/java/core/basesyntax/db/StorageTest.java
```diff
@@ -0,0 +1,46 @@
...
+        assertInstanceOf(HashMap.class, Storage.getFruitStorage(),
+                "Expected instance of HashMap");
+    }
+
+    @Test
+    void storage_acceptsStringAndInteger_Ok() {
```
nowickimj, line 33: Similar case here - you're checking something trivial. 
If your storage implements some specific add/delete operation, it would be reasonable to compare the state before & after the operation, but we're dealing with a regular hash map here.

src/test/java/core/basesyntax/service/impl/DataConverterServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.NoSuchElementException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class DataConverterServiceImplTest {
+    private static final List<String> operationsValid1 =
+            readFileContents("src/test/resources/operations_valid_1.csv");
+    private static final List<String> operationsValid2 =
```
nowickimj, line 21: also, if `operationsValid1` and `operationsValid2` are defined for specific cases, better use explicit name, like "OPERATIONS_SINGLE_FRUIT" & "OPERATIONS_MANY_FRUITS"

src/test/java/core/basesyntax/service/impl/DataConverterServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+            readFileContents("src/test/resources/operations_random_text.csv");
+    private static DataConverterService dataConverter = new DataConverterServiceImpl();
+
+    private FruitTransaction transaction1;
+    private FruitTransaction transaction2;
+    private FruitTransaction transaction3;
+    private FruitTransaction transaction4;
+    private List<FruitTransaction> transactions;
+
+    @BeforeEach
```
nowickimj, line 39: I think you should make `FruitTransaction` objects local - they're only used in few tests, so  the usage of `@BeforeEach` may be misleading.

nowickimj (changes requested):

PR #910, opened 2024-02-22
Implemented tests

shumYevh:


okuzan (changes requested):
Please make your methods conform to the described naming convention, they should look smth like this `read_nonexistentFile_ok` 
image

src/test/java/service/impl/FruitTransactionParserTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.List;
+import model.FruitTransaction;
+import org.junit.jupiter.api.Test;
+import service.Parser;
+
+class FruitTransactionParserTest {
+
```
okuzan, line 12:
```suggestion
```

src/test/java/service/impl/CsvFruitReaderTest.java
```diff
@@ -0,0 +1,38 @@
...
+            "    b,apple,100",
+            "    s,banana,100",
+            "    p,banana,13",
+            "    r,apple,10",
+            "    p,apple,20",
+            "    p,banana,5",
+            "    s,banana,50");
+
+    @Test
+    void read_textFromFile_Ok() {
```
okuzan, line 24:
```suggestion
    void read_TextFromFile_Ok() {
```
either always use lowercase after underscore, or always use uppercase

src/test/java/service/impl/FruitOperationPerformerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import strategy.operation.FruitReturnOperation;
+import strategy.operation.FruitSupplyOperation;
+import strategy.operation.OperationHandler;
+
+class FruitOperationPerformerTest {
+    private static FruitStorage FRUIT_STORAGE;
+    private static StorageDao<String,Integer> STORAGE_HANDLER;
+    private static Map<FruitTransaction.Operation,
+            OperationHandler<String, Integer>> OPERATION_HANDLER_MAP;
+    private static TransactionStrategy MAKE_TRANSACTION;
```
okuzan, line 29: They are not declared as final; why are they named in uppercase as if they were constants? According to the Google Java Style Guide, static final members should only be in uppercase if they are constants, which implies deep immutability

okuzan (changes requested):
Good job on your tests, some additional code style changes suggested

olekskov (approved):
Good job

PR #912, opened 2024-02-29
dsumaroka/jv-fruit-shop-tests

dsumaroka:


okuzan (changes requested):
These and DaoImpl should be tested, each in its own separate file
image
image

okuzan (changes requested):
I meant there should be a separate test class for each of them

dsumaroka:
Ок)

Пт, 1 марта 2024 г. в 18:12, Oleh Kuzan ***@***.***>:

src/test/java/core/basesyntax/FileReaderTest.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
```
olekskov, line 1: It's a good practice to create same folders structure for tests as for classes they test.
If you FileReaderImpl is in folder `core.basesyntax.service.impl` then test should be also in forder 'service.impl' in test package.
Ans similar for other classes

olekskov (approved):
Good job overall, approved.
Please pay attention to comment below for further

dsumaroka:
Ok. Thank you

Сб, 2 марта 2024 г. в 13:37, Oleksandr Kovalenko ***@***.***>:

PR #922, opened 2024-03-16
added hw_solution

SkillfulDev:


EduarfF (approved):
Great job

src/main/java/core/basesyntax/factory/impl/OperationHandlerFactoryImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class OperationHandlerFactoryImpl implements OperationHandlerFactory {
+    private final Map<Operation, OperationHandler> operations = new HashMap<>();
+
+    public OperationHandlerFactoryImpl() {
+        operations.put(Operation.BALANCE, new BalanceStrategyImpl());
+        operations.put(Operation.RETURN, new ReturnStrategyImpl());
+        operations.put(Operation.PURCHASE, new PurchaseStrategyImpl());
+        operations.put(Operation.SUPPLY, new SupplyStrategyImpl());
+    }
```
karma-o, line 21: Just my thoughts on the original implementation of the shop, you do not have to fix the comments regarding it, unless it is necessary for writing proper tests.
I get the idea of a factory, that it produces a map for the strategy, but we have a bit of a mix in responsibilities here:
* producing the handlers
* getting the one we need based on the Operation(strategy) 

To fix that I propose to receive that map of handlers in the constructor to move the fabrication responsibility outside, and then it is not a factory anymore, but just a strategy that could be named accordingly

src/test/java/core/basesyntax/service/impl/FruitMapperImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+        inputData.add(VALID_FIRST_ROW);
+        inputData.add(VALID_SECOND_ROW);
+        int quantityForBanana = 20;
+        int quantityForApple = 100;
+        List<Fruit> expected = List.of(
+                new Fruit(FRUIT_BANANA, Operation.BALANCE, quantityForBanana),
+                new Fruit(FRUIT_APPLE, Operation.SUPPLY, quantityForApple)
+        );
+        List<Fruit> actual = fruitMapper.convert(inputData);
+        assertEquals(expected, actual, "Fruits should be correctly mapped from input data.");
```
karma-o, line 42: Just to expand your knowledge a bit, there is a method and various constructions around it like `assertThat(actual).using...().ignoring...().isEqualTo(expected)` that is very useful in a more complex comarisions, which provides you with a more flexible approaches to asserts, like ignoring collection order, using of recursive field by field comparison ignoring overrided equals or some fields during the comparison, or in case you are using some other classes and there is no equals. Mostly useful in the assertion of collections but also has many overloaded variants for different types

src/test/java/core/basesyntax/service/impl/ProcessDataServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+                new Fruit(FRUIT_APPLE, Operation.PURCHASE, 10));
+        RuntimeException exception = assertThrows(RuntimeException.class,
+                () -> processDataService.processData(fruits));
+        assertEquals("Total amount for apple cannot be negative",
+                exception.getMessage(), "Exception message should match.");
+    }
+
+    @Test
+    void emptyListNoChangeToStorage_processData_ok() {
+        processDataService.processData(List.of());
```
karma-o, line 62:
```suggestion
        processDataService.processData(Collections.emptyList());
```

src/test/java/core/basesyntax/service/impl/ReaderFromFileImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class ReaderFromFileImplTest {
+    private static final String VALID_INPUT_FILE = "src/main/resources/inputData.csv";
```
karma-o, line 17: We should not use main resources in the tests, create a duplicate to separate them. 
Exception can be a usage of SQL scripts in main to set up the test DB but we will talk about that later on

src/test/java/core/basesyntax/service/impl/ReaderFromFileImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+        Path emptyFile = Files.createFile(temporaryDirectory.resolve(EMPTY_FILE));
+        List<String> actualLines = reader.readFromFile(emptyFile.toString());
+        assertTrue(actualLines.isEmpty(),
+                "The read data should be an empty list for an empty file.");
+    }
+
+    @Test
+    void invalidInputFile_readFromFile_notOk() {
+        String invalidInputFile = "src/main/resources/outputDat.csv";
+        RuntimeException runtimeException = assertThrows(RuntimeException.class,
```
karma-o, line 57: can we be more specific on the type of exception here? if we throw it by ourselves, we should replace it with a custom one in the implementation
And let's name this var something like `thrownException` or `actualException` or `actual`

src/test/java/core/basesyntax/service/impl/WriterImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+        String actualReport = Files.readString(outputPath);
+        assertEquals(expectedReport, actualReport,
+                "The written content should match the expected report.");
+    }
+
+    @Test
+    void invalidPath_write_ThrowsRuntimeException() {
+        String invalidPath = temporaryDirectory.resolve(
+                "nonexistent/directory/outputData.csv").toString();
+        Exception exception = assertThrows(RuntimeException.class,
```
karma-o, line 41: same thing about the type of the exception and var name

karma-o (changes requested):
Some fixes need to be done, but looks very good! :+1:  
I have also left some comments to the logic itself, most of them are optional so feel free to ignore them

src/main/java/core/basesyntax/strategy/OperationHandler.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.strategy;
+
+public interface OperationHandler {
```
karma-o, line 3: BTW, forgot about that in the first look, we also need to cover handlers with tests, and probably even the strategy class that provides handlers

src/main/java/core/basesyntax/strategy/impl/PurchaseStrategyImpl.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.function.IntUnaryOperator;
+
+public class PurchaseStrategyImpl implements OperationHandler {
+    private final IntUnaryOperator operation = value -> -value;
+
+    @Override
+    public int calculate(int amount) {
+        return operation.applyAsInt(amount);
+    }
+}
```
karma-o, line 13: would also be nice to add a check if we have enough fruit amount in the storage to sell it, and throw ex if we are out of stock on attempt to buy it, and a test for it

src/main/java/core/basesyntax/factory/impl/OperationHandlerFactoryImpl.java
SkillfulDev: I suppose It makes sense, need to process this info

src/test/java/core/basesyntax/service/impl/FruitMapperImplTest.java
SkillfulDev: Thank you I`ll explore this question

src/test/java/core/basesyntax/service/impl/ReaderFromFileImplTest.java
SkillfulDev: but we don`t have a custom exception. Should I create one?

src/test/java/core/basesyntax/service/impl/WriterImplTest.java
SkillfulDev: I realize the core concept of custom exception, I need to have my own, but should I  now change my source code?

src/main/java/core/basesyntax/strategy/impl/PurchaseStrategyImpl.java
SkillfulDev: for this case we have.

image and test 
image
Does the strategy have to know something to the object ?

PR #934, opened 2024-03-22
My fruit shop test

petro2061:
In the task, unit tests are written that check the implementation of reading data from a file, processing data using handlers and writing the result to a file.

### DIFF
--- a/src/test/java/core/basesyntax/service/impl/FruitTransactionParserImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/FruitTransactionParserImplTest.java
@@ -1,0 +1,89 @@
+package core.basesyntax.service.impl;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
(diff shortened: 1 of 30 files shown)


src/test/java/core/basesyntax/service/impl/FruitTransactionParserImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+    void converterToTransactions_checkedCorrectSkipFirstLineInInputList_Ok() {
+        String firstLine = "type,fruit,quantity";
+        List<FruitTransaction> fruitTransactions
+                = testTransactionParser.converterToTransactions(inputTestList);
+        boolean contains = fruitTransactions.contains(firstLine);
+        assertFalse(contains, "Can not skip firstLine in input data");
+    }
+
+    @Test
+    void converterToTransactions_inputNegativeQuantityPerOperation_NotOk() {
```
okuzan, line 44: it's a bit weird that after first underscore you use lowercase, after second - uppercase, please unify, let's always use lowercase

okuzan (changes requested):
Good job! left some code style comments

petro2061:
Fixed problem with naming methods also add new test case

olekskov (approved):

PR #939, opened 2024-03-22
Fruit shop solution tests

Nikname2303:


src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+import core.basesyntax.strategy.OperationHandler;
+import core.basesyntax.strategy.impl.BalanceHandler;
+import core.basesyntax.strategy.impl.PurchaseHandler;
+import core.basesyntax.strategy.impl.ReturnHandler;
+import core.basesyntax.strategy.impl.SupplyHandler;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Assertions;
```
VolandevlodD, line 15: To enhance readability and conciseness, consider using static imports for assertions used in the test classes.

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+        List<String> actual = fileReaderCsv.read(DEFAULT_VALID_FILE);
+        Assertions.assertEquals(actual, fileReaderCsv.read(DEFAULT_VALID_FILE));
+    }
+
+    @Test
+    void read_invalidPath_notOk() {
+        Throwable exception = Assertions.assertThrows(InvalidDataException.class, () -> {
+            fileReaderCsv.read(INVALID_PATH_TO_FILE);
+        });
+        Assertions.assertEquals("Can't read file " + INVALID_PATH_TO_FILE, exception.getMessage());
```
VolandevlodD, line 40: After utilizing static imports, the code can be simplified to:
```suggestion
        assertThrows(InvalidDataException.class,() ->
                fileReaderCsv.read(INVALID_PATH_TO_FILE),"Can't read file " + INVALID_PATH_TO_FILE);
```
Since assertThrows method is overloaded.

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    void parse_validData_Ok() {
+        List<FruitTransaction> actual = fruitTransactionParser.parse(fileReaderCsv
+                .read(DEFAULT_VALID_FILE));
+        List<FruitTransaction> expected = new ArrayList<>();
+        Assertions.assertSame(expected.getClass(), actual.getClass());
+    }
```
VolandevlodD, line 49: Знімок екрана 2024-03-23 211033
Looks like you just checked that both are instances of ArrayList.

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    void parse_invalidQuantity_notOk() {
```
VolandevlodD, line 52: Since you're throwing your custom exception here, I think it would be good practice to clarify that you're expecting to receive this exception in your method name.
```suggestion
    void parse_invalidQuantity_ThrowsInvalidDataException() {
```

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+        Throwable exception = Assertions.assertThrows(InvalidDataException.class, () -> {
+            fruitTransactionParser.parse(fileReaderCsv.read(INVALID_QUANTITY_IN_FILE));
+        });
+        Assertions.assertEquals("Input quantity is not valid", exception.getMessage());
```
VolandevlodD, line 56: You can simplify the code like this:
```suggestion
      assertThrows(InvalidDataException.class, () ->
        fruitTransactionParser.parse(fileReaderCsv.read(INVALID_QUANTITY_IN_FILE)),
        "Input quantity is not valid");`
```
It`s up to you)

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+        });
+        Assertions.assertEquals("Fruit does not exist", exception.getMessage());
+    }
+
+    @Test
+    void generateReport_validData_Ok() {
+        String actual = reportGenerator.generateReport(Storage.fruits);
+        String expected = "expected";
+        Assertions.assertSame(expected.getClass(), actual.getClass());
+    }
```
VolandevlodD, line 80: What are you checking in this test?

src/test/java/core/basesyntax/service/impl/FileReaderCsvTest.java
```diff
@@ -0,0 +1,95 @@
...
+        strategyMap.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler());
+        strategyMap.put(FruitTransaction.Operation.RETURN, new ReturnHandler());
+        strategyMap.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler());
+        TransactionExecutor transactionExecutor = new TransactionExecutor(strategyMap);
+        transactionExecutor.execute(fruitTransactionParser.parse(fileReaderCsv
+                .read(DEFAULT_VALID_FILE)));
+        Assertions.assertEquals(transactionExecutor.getStrategy()
+                .get(FruitTransaction.Operation.BALANCE).getClass(), BalanceHandler.class);
+    }
+}
```
VolandevlodD, line 95: In my opinion, it's not advisable to use methods from your application directly in your test methods. Instead, it's better to utilize basic API methods for testing purposes.

VolandevlodD (changes requested):
Please review the checklist and consider separating the test logic into different classes for improved organization and clarity.

andrii-hoienko (changes requested):
Please check common mistakes and consider student comments

src/test/java/core/basesyntax/service/impl/TransactionExecutorTest.java
```diff
@@ -0,0 +1,33 @@
...
+import core.basesyntax.strategy.OperationHandler;
+import core.basesyntax.strategy.impl.BalanceHandler;
+import core.basesyntax.strategy.impl.PurchaseHandler;
+import core.basesyntax.strategy.impl.ReturnHandler;
+import core.basesyntax.strategy.impl.SupplyHandler;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class TransactionExecutorTest {
```
karma-o, line 15: we should also test `OperationStrategy` and each `Handler` in their respective test classes

src/test/java/core/basesyntax/service/impl/TransactionExecutorTest.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final String DEFAULT_VALID_FILE = "src/main/resources/example.csv";
+    private FruitTransactionParserImpl fruitTransactionParser = new FruitTransactionParserImpl();
+    private FileReaderCsv fileReaderCsv = new FileReaderCsv();
+
+    @Test
+    void execute_validData_Ok() {
```
karma-o, line 21: this is an end-to-end test, which is nice to have, but we also need unit tests to be able to understand which exact part is broken, in case it is. 
So the unit test for `TransactionExecutor` would be to test if it will process all transactions, or maybe just some of them, and test it's behavior with different params like empty or null list

src/test/java/core/basesyntax/service/impl/ReportGeneratorTest.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ReportGeneratorTest {
+    private ReportGenerator reportGenerator = new ReportGenerator();
+
+    @Test
+    void generateReport_validData_Ok() {
+        String actual = reportGenerator.generateReport(Storage.fruits);
+        assertEquals(String.class, actual.getClass());
+    }
```
karma-o, line 15: add a case with a non-empty storage, because this one does not really test a valid data

src/test/java/core/basesyntax/service/impl/TransactionExecutorTest.java
```diff
@@ -0,0 +1,33 @@
...
+        Map<FruitTransaction.Operation, OperationHandler> strategyMap = new HashMap<>();
+        strategyMap.put(FruitTransaction.Operation.BALANCE, new BalanceHandler());
+        strategyMap.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler());
+        strategyMap.put(FruitTransaction.Operation.RETURN, new ReturnHandler());
+        strategyMap.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler());
+        TransactionExecutor transactionExecutor = new TransactionExecutor(strategyMap);
+        transactionExecutor.execute(fruitTransactionParser.parse(fileReaderCsv
+                .read(DEFAULT_VALID_FILE)));
+        assertEquals(transactionExecutor.getStrategy()
+                .get(FruitTransaction.Operation.BALANCE).getClass(), BalanceHandler.class);
```
karma-o, line 31: this assert is redundant and does not have anything to do with the case of testing how `execute` method works, and in general this `getStrategy()` method is redundant, I'd say we never should have a situation where we need to get the service of some other service.
Instead, we should assert storage data here, or write a report and check its file contents if you really want to commit to this end-to-end test

karma-o (changes requested):

PR #949, opened 2024-03-26
Jv fruit shop tests hw solution

JohnGrey17:


bodya4243 (approved):
Build a project

karma-o (changes requested):
Build failed, we should fix that first

JohnGrey17:
> Build failed, we should fix that first

I did not add all committed classes that why got an issue

andrii-hoienko (approved):
Good job!

PR #950, opened 2024-03-27
feat: initial commit

rel1c-hub:
build success

### DIFF
--- a/src/main/java/core/basesyntax/service/strategy/impl/OperationStrategyImpl.java
+++ b/src/main/java/core/basesyntax/service/strategy/impl/OperationStrategyImpl.java
@@ -1,0 +1,21 @@
+package core.basesyntax.service.strategy.impl;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.operations.OperationHandler;
+import core.basesyntax.service.strategy.Operation;
+import core.basesyntax.service.strategy.OperationStrategy;
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private final Map<Operation, OperationHandler> strategyMap;
+
+    public OperationStrategyImpl(Map<Operation, OperationHandler> strategyMap) {
+        this.strategyMap = strategyMap;
+    }
+
+    @Override
+    public OperationHandler get(FruitTransactionDto dto) {
+        Operation operation = Operation.validateOperation(dto.operation().getCode());
+        return strategyMap.get(operation);
+    }
+}
--- a/src/main/java/core/basesyntax/service/strategy/impl/TransactionProcessorImpl.java
+++ b/src/main/java/core/basesyntax/service/strategy/impl/TransactionProcessorImpl.java
@@ -1,0 +1,22 @@
+package core.basesyntax.service.strategy.impl;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.operations.OperationHandler;
+import core.basesyntax.service.strategy.OperationStrategy;
+import core.basesyntax.service.strategy.TransactionProcessor;
+import java.util.List;
+
+public class TransactionProcessorImpl implements TransactionProcessor {
+    private OperationStrategy operationStrategy;
+
+    public TransactionProcessorImpl(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    public void process(List<FruitTransactionDto> transactions) {
+        for (var transaction : transactions) {
+            OperationHandler handler = operationStrategy.get(transaction);
+            handler.apply(transaction);
+        }
+    }
+}
--- a/src/test/java/core/basesyntax/service/impl/FileWriterImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/FileWriterImplTest.java
@@ -1,0 +1,54 @@
+package core.basesyntax.service.impl;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+import core.basesyntax.service.operations.report.FileWriter;
+import core.basesyntax.service.operations.report.impl.FileWriterImpl;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class FileWriterImplTest {
+    private static final String TEST_FILE_PATH = "src/test/resources/test_report.txt";
+    private static final String TEST_CONTENT = "This is a test report content.";
+    private FileWriter fileWriter;
+
+    @BeforeEach
+    void setUp() {
+        fileWriter = new FileWriterImpl();
+    }
+
+    @AfterEach
+    void tearDown() throws IOException {
+        Files.deleteIfExists(Path.of(TEST_FILE_PATH));
+    }
+
+    @Test
+    void writeDataToFile_ValidFilePath_ok() throws IOException {
+        fileWriter.writeDataToFile(TEST_CONTENT, TEST_FILE_PATH);
+        File file = new File(TEST_FILE_PATH);
+        assertTrue(file.exists());
+        List<String> actualLines = Files.readAllLines(Path.of(TEST_FILE_PATH));
+        assertEquals(1, actualLines.size());
+        assertEquals(TEST_CONTENT, actualLines.get(0));
+    }
+
+    @Test
+    void writeDataToFile_NullFilePath_notOk() {
+        assertThrows(NullPointerException.class, () ->
+                fileWriter.writeDataToFile(TEST_CONTENT, null));
+    }
+
+    @Test
+    void writeDataToFile_NullContent_notOk() {
+        assertThrows(RuntimeException.class, () ->
+                fileWriter.writeDataToFile(null, TEST_FILE_PATH));
+    }
+}
--- a/src/test/java/core/basesyntax/service/impl/FruitDataParserTest.java
+++ b/src/test/java/core/basesyntax/service/impl/FruitDataParserTest.java
@@ -1,0 +1,67 @@
+package core.basesyntax.service.impl;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+import core.basesyntax.FruitDataParser;
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.DataParser;
+import core.basesyntax.service.strategy.Operation;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class FruitDataParserTest {
+    private DataParser<FruitTransactionDto> dataParser;
+
+    @BeforeEach
+    void setUp() {
+        dataParser = new FruitDataParser();
+    }
+
+    @Test
+    void parse_ValidData_Ok() {
+        List<String> rawData = List.of(
+                "type,fruit,quantity",
+                "b,banana,20",
+                "p,apple,50",
+                "s,orange,30"
+        );
+
+        List<FruitTransactionDto> expected = List.of(
+                new FruitTransactionDto(Operation.BALANCE, "banana", 20),
+                new FruitTransactionDto(Operation.PURCHASE, "apple", 50),
+                new FruitTransactionDto(Operation.SUPPLY, "orange", 30)
+        );
+
+        List<FruitTransactionDto> actual = dataParser.parse(rawData);
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void parse_InvalidData_notOk() {
+        List<String> rawData = List.of(
+                "type,fruit,quantity",
+                "b,banana",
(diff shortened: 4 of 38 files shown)


src/test/java/core/basesyntax/service/impl/FileWriterImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+        assertTrue(file.exists());
+        List<String> actualLines = Files.readAllLines(Path.of(TEST_FILE_PATH));
+        assertEquals(1, actualLines.size());
+        assertEquals(TEST_CONTENT, actualLines.get(0));
+    }
+
+    @Test
+    void writeDataToFile_NullFilePath_notOk() {
+        assertThrows(NullPointerException.class, () ->
+                fileWriter.writeDataToFile(TEST_CONTENT, null));
```
karma-o, line 46: we should add a validation of the file path to the impl class, because asserting an NPE is a bad practice since you can not guarantee that this NPE will be thrown exactly because of the reason ou think it is

src/test/java/core/basesyntax/service/impl/FileWriterImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    void writeDataToFile_NullContent_notOk() {
+        assertThrows(RuntimeException.class, () ->
```
karma-o, line 51: Let's replace all raw Runtime exceptions with a custom ones, in both impl and test methods, because NPE for example is also a Runtime, and so we expect any exception here really, which  is not good
And also assert the exception message

src/test/java/core/basesyntax/service/impl/FruitDataParserTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    void parse_InvalidData_notOk() {
+        List<String> rawData = List.of(
+                "type,fruit,quantity",
+                "b,banana",
+                "p,apple,50",
+                "s,orange,30"
+        );
```
karma-o, line 48: do you see the code duplication with the case above? let's make it a class var and reinitialize it in the beforeEach

src/test/java/core/basesyntax/service/impl/FruitDataParserTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        assertThrows(IllegalArgumentException.class, () ->
+                dataParser.parse(rawData));
+    }
+
+    @Test
+    void parse_EmptyData_ReturnsEmptyList() {
+        List<String> rawData = List.of();
```
karma-o, line 56: just a minor thing
```suggestion
        List<String> rawData = Collections.emptyList()
```

src/test/java/core/basesyntax/service/impl/FruitDataParserTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        List<FruitTransactionDto> result = dataParser.parse(rawData);
+        assertEquals(0, result.size());
+    }
+
+    @Test
+    void parse_NullData_notOk() {
+        assertThrows(NullPointerException.class, () ->
```
karma-o, line 64: same here, add null check to the impl class

src/test/java/core/basesyntax/service/impl/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+        balanceOperationHandler = new BalanceOperationHandler(storage);
+    }
+
+    @Test
+    public void apply_AddToFruitStorage_ok() {
+        FruitTransactionDto dto = new FruitTransactionDto(Operation.BALANCE, FRUIT_NAME,
+                BALANCE_QUANTITY);
+        balanceOperationHandler.apply(dto);
+        Map<Fruit, Integer> fruits = storage.getFruits();
+        assertEquals(BALANCE_QUANTITY, fruits.getOrDefault(new Fruit(FRUIT_NAME), 0));
```
karma-o, line 33: why do we need this default? and let's first put it into the variable named `actual` and then assert it

src/test/java/core/basesyntax/service/impl/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        storage = new Storage();
+        balanceOperationHandler = new BalanceOperationHandler(storage);
+    }
+
+    @Test
+    public void apply_AddToFruitStorage_ok() {
+        FruitTransactionDto dto = new FruitTransactionDto(Operation.BALANCE, FRUIT_NAME,
```
karma-o, line 29: we can have lots of DTO classes across our application, but a `transaction` is more clear and unique name for it
```suggestion
        FruitTransactionDto transaction = new FruitTransactionDto(Operation.BALANCE, FRUIT_NAME,
```

src/test/java/core/basesyntax/service/impl/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+                BALANCE_QUANTITY);
+        balanceOperationHandler.apply(dto);
+        Map<Fruit, Integer> fruits = storage.getFruits();
+        assertEquals(BALANCE_QUANTITY, fruits.getOrDefault(new Fruit(FRUIT_NAME), 0));
+    }
+
+    @Test
+    public void apply_AddToExistingFruit_ok() {
+        int initialQuantity = 0;
```
karma-o, line 38: make it a constant

src/test/java/core/basesyntax/service/impl/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+        storage.addFruit(new Fruit(FRUIT_NAME), initialQuantity);
+        FruitTransactionDto dto = new FruitTransactionDto(Operation.BALANCE, FRUIT_NAME,
```
karma-o, line 40: same here and in other places about the var name

src/test/java/core/basesyntax/service/impl/strategy/impl/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+                BALANCE_QUANTITY);
+        balanceOperationHandler.apply(dto);
+        Map<Fruit, Integer> fruits = storage.getFruits();
+        assertEquals(initialQuantity + BALANCE_QUANTITY,
+                fruits.getOrDefault(new Fruit(FRUIT_NAME), 0));
```
karma-o, line 45: same here, you do not need this default value

src/test/java/core/basesyntax/service/impl/strategy/impl/PurchaseOperationHandlerTest.java
```diff
@@ -0,0 +1,53 @@
...
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.service.operations.impl.PurchaseOperationHandler;
+import core.basesyntax.service.strategy.Operation;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class PurchaseOperationHandlerTest {
```
karma-o, line 16: same as in the previous comments for this test class

src/test/java/core/basesyntax/service/impl/strategy/impl/ReturnOperationHandlerTest.java
```diff
@@ -0,0 +1,52 @@
...
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.service.operations.impl.ReturnOperationHandler;
+import core.basesyntax.service.strategy.Operation;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ReturnOperationHandlerTest {
```
karma-o, line 15: and this

src/test/java/core/basesyntax/service/impl/strategy/impl/SupplyOperationHandlerTest.java
```diff
@@ -0,0 +1,53 @@
...
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.service.operations.impl.SupplyOperationHandler;
+import core.basesyntax.service.strategy.Operation;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class SupplyOperationHandlerTest {
```
karma-o, line 15: and here

src/main/java/core/basesyntax/service/strategy/impl/TransactionProcessorImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.strategy.impl;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.operations.OperationHandler;
+import core.basesyntax.service.strategy.OperationStrategy;
+import core.basesyntax.service.strategy.TransactionProcessor;
+import java.util.List;
+
+public class TransactionProcessorImpl implements TransactionProcessor {
```
karma-o, line 9: also we need to test this class

src/main/java/core/basesyntax/service/strategy/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax.service.strategy.impl;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.operations.OperationHandler;
+import core.basesyntax.service.strategy.Operation;
+import core.basesyntax.service.strategy.OperationStrategy;
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
```
karma-o, line 9: and this one too

karma-o (changes requested):

src/test/java/core/basesyntax/service/strategy/impl/TransactionProcessorImplTest.java
```diff
@@ -0,0 +1,40 @@
...
+        operationStrategy = new OperationStrategyImpl(strategyMap);
+        transactionProcessor = new TransactionProcessorImpl(operationStrategy);
+    }
+
+    @Test
+    public void process_NullList_notOk() {
+        CustomException customException = assertThrows(CustomException.class,
+                () -> transactionProcessor.process(null));
+        assertEquals("Transaction list is null", customException.getMessage());
+    }
```
karma-o, line 39: we can also add a valid case and assert storage elements there

src/test/java/core/basesyntax/service/strategy/impl/TransactionProcessorImplTest.java
```diff
@@ -0,0 +1,40 @@
...
+    private TransactionProcessorImpl transactionProcessor;
+    private OperationStrategy operationStrategy;
+
+    @BeforeEach
+    public void setUp() {
+        Map<Operation, OperationHandler> strategyMap = Map.of(
+                Operation.BALANCE, new BalanceOperationHandler(new Storage()),
+                Operation.PURCHASE, new PurchaseOperationHandler(new Storage()),
+                Operation.RETURN, new ReturnOperationHandler(new Storage()),
+                Operation.SUPPLY, new SupplyOperationHandler(new Storage()));
```
karma-o, line 29: we reuse a single storage instance for all handlers

src/main/java/exception/CustomException.java
```diff
@@ -0,0 +1,11 @@
+package exception;
+
+public class CustomException extends RuntimeException {
```
karma-o, line 3: I did not mean literally "Custorm" 😅 
we should have a proper name for it, depending on the context of the place where it is thrown, and 1 exception most likely won't be enough, because we have a big application here with lots of different logic, and each case should be considered on it's own, for example:
`InvalidInputDataException`, `NotEnoghFruitsException`, or some places can reuse the ones from jdk, like `IllegalArgumentException`, Overall it is fine to reuse exceptions but only if they are fitting in the context, and they should be named properly

src/test/java/core/basesyntax/service/strategy/impl/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+        operationStrategy = new OperationStrategyImpl(strategyMap);
+    }
+
+    @Test
+    public void get_NullOperation_noOk() {
+        fruitTransactionDto = new FruitTransactionDto(null, "apple", 5);
+        CustomException customException = assertThrows(CustomException.class,
+                () -> operationStrategy.get(fruitTransactionDto));
+        assertEquals("Operation cannot be null", customException.getMessage());
+    }
```
karma-o, line 42: add a valid case for getting a handler and validating if it has returned the proper type

karma-o (approved):
Approved but left several important comments, please take a look!

PR #958, opened 2024-04-08
tests were written and added hashCode and equals methods in FruitTran…

nastasiiaKulinich:
…saction class

### DIFF
--- a/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
@@ -1,0 +1,50 @@
+package core.basesyntax.service.impl;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.service.WriterService;
+import java.util.ArrayList;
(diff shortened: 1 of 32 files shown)


src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertThrows(RuntimeException.class, () -> writerService.writeToFile(null, lines));
+    }
+
+    @Test
+    void writeToFile_validDataAndPath_ok() {
+        List<String> lines = new ArrayList<>();
+        lines.add("fruit,quantity");
+        lines.add(System.lineSeparator() + "banana,20");
+        lines.add(System.lineSeparator() + "apple,10");
+        writerService.writeToFile(VALID_FILE_NAME, lines);
```
olekskov, line 48: here we also should check file content as if method haven't thrown exception it doesn't mean it works as intended

olekskov (approved):
Good job, approved.
Please pay attention to comment below for further

PR #969, opened 2024-04-20
project is ready for the code review

misterSokor:
The project is ready, but I can't understand why when I run the command 'mvn clean verify' in the terminal, it says that I did not meet the project requirements and have covered just 0.7%, but I need to cover at least 0.8%. However, when I run it with coverage in the project, it shows that I've covered 100%. What could be the reason? Thank you!

### DIFF
--- a/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
@@ -10,11 +10,16 @@ public class ReportServiceImpl implements ReportService {
 
     @Override
     public List<String> generateReport(Map<String, Integer> fruitCounts) {
-        List<String> report = new ArrayList<>();
-        report.add(CATEGORIES);
-        for (Map.Entry<String, Integer> entry : fruitCounts.entrySet()) {
-            report.add(entry.getKey() + "," + entry.getValue());
+        if (fruitCounts == null) {
+            throw new NullPointerException("Can't generate report for null data");
+        } else {
+            List<String> report = new ArrayList<>();
+            report.add(CATEGORIES);
+            for (Map.Entry<String, Integer> entry : fruitCounts.entrySet()) {
+                report.add(entry.getKey() + "," + entry.getValue());
+            }
+            return report;
         }
-        return report;
     }
 }
+
--- a/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
@@ -1,0 +1,20 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReportService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    private static final String CATEGORIES = "fruit,quantity";
+
+    @Override
+    public List<String> generateReport(Map<String, Integer> fruitCounts) {
+        List<String> report = new ArrayList<>();
+        report.add(CATEGORIES);
+        for (Map.Entry<String, Integer> entry : fruitCounts.entrySet()) {
+            report.add(entry.getKey() + "," + entry.getValue());
+        }
+        return report;
+    }
+}
--- a/src/main/java/core/basesyntax/service/impl/TransactionProcessorServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/TransactionProcessorServiceImpl.java
@@ -1,0 +1,43 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.TransactionProcessorService;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.List;
+import java.util.Map;
+
+public class TransactionProcessorServiceImpl implements TransactionProcessorService {
+    private Map<String, Integer> fruitCounts;
+    private final Map<FruitTransaction.Operation, OperationStrategy> strategies;
+
+    public TransactionProcessorServiceImpl(Map<FruitTransaction.Operation,
+            OperationStrategy> strategies) {
+        this.strategies = strategies;
+    }
+
+    @Override
+    public Map<String, Integer> processTransaction(List<FruitTransaction> transactions) {
+        fruitCounts = Storage.getFruits();
+        if (transactions != null) {
+            for (FruitTransaction transaction : transactions) {
+                processSingleTransaction(transaction);
+            }
+        }
+        return fruitCounts;
+    }
+
+    public void processSingleTransaction(FruitTransaction transaction) {
+        FruitTransaction.Operation typeOfOperation = transaction.getType();
+        String fruit = transaction.getFruit();
+        int quantity = transaction.getQuantity();
+
+        OperationStrategy strategy = strategies.get(typeOfOperation);
+        if (strategy != null) {
+            strategy.apply(fruitCounts, fruit, quantity);
+
+        } else {
+            System.out.println("Invalid operation type: " + typeOfOperation);
+        }
+    }
+}
--- a/src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
@@ -1,0 +1,51 @@
+package core.basesyntax.service.impl;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ParseServiceImplTest {
+    private ParseServiceImpl parseService;
+
+    @BeforeEach
+    void setUp() {
+        parseService = new ParseServiceImpl();
+    }
+
+    @Test
+    void parseFromString_emptyList() {
+        List<String> transactionLine = List.of();
+        assertTrue(parseService.parseFromString(transactionLine).isEmpty());
+    }
+
+    @Test
+    void parseFromString_validData() {
+        List<String> transactionLine = List.of("b,banana,10", "s,apple,5");
+        List<FruitTransaction> result =
+                parseService.parseFromString(transactionLine);
+        assertEquals(2, result.size());
+        assertEquals(FruitTransaction.Operation.BALANCE,
+                result.get(0).getType());
+        assertEquals("banana", result.get(0).getFruit());
+        assertEquals(10, result.get(0).getQuantity());
+        assertEquals(3, result.get(0).getClass().getDeclaredFields().length);
+    }
+
+    @Test
+    void parseFromString_invalidData() {
+        List<String> transactionLine = List.of("b, banana");
+        Exception exception =
+                assertThrows(IllegalArgumentException.class,
+                        () -> parseService.parseFromString(transactionLine));
+
+        String expectedMessage =
+                "Invalid transaction format: " + transactionLine;
+        String actualMessage = exception.getMessage();
+        assertTrue(actualMessage.contains(expectedMessage));
+    }
+}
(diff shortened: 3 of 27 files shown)


src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+class WriterServiceImplTest {
+    private WriterService writerService;
+    private String filePath;
+    private List<String> originalContent;
+
+    @BeforeEach
+    void setUp() throws IOException {
+        writerService = new WriterServiceImpl();
+        filePath = "src/main/resources/output.csv";
```
okuzan, line 25: file paths like this are virtually always constants

src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+class ParseServiceImplTest {
+    private ParseServiceImpl parseService;
+
+    @BeforeEach
+    void setUp() {
+        parseService = new ParseServiceImpl();
+    }
+
+    @Test
+    void parseFromString_emptyList() {
```
okuzan, line 21: please adhere to the established naming convention
add _ok or _throwsSomeException or just _notOk, third part of method name should be expected result, it will boost readability

src/test/java/core/basesyntax/service/impl/ParseServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+    void parseFromString_validData() {
+        List<String> transactionLine = List.of("b,banana,10", "s,apple,5");
+        List<FruitTransaction> result =
+                parseService.parseFromString(transactionLine);
+        assertEquals(2, result.size());
+        assertEquals(FruitTransaction.Operation.BALANCE,
+                result.get(0).getType());
+        assertEquals("banana", result.get(0).getFruit());
+        assertEquals(10, result.get(0).getQuantity());
+        assertEquals(3, result.get(0).getClass().getDeclaredFields().length);
```
okuzan, line 36: looks a bit messy
i would just create the expected transactions myself and assert against them

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @BeforeEach
+    void setUp() throws IOException {
+        readerService = new ReaderServiceImpl();
+    }
+
+    @Test
+    void fileNotFound() {
+        Exception exception = assertThrows(RuntimeException.class,
+                () -> readerService.readFromFilesContents("src/main/resources/missing_file.txt"));
+        String expectedMessage = "Can't read from the file: src/main/resources/missing_file.txt";
```
okuzan, line 25: When you find yourself reusing a string, it's always a good idea to extract it to a constant. This way, you will only ever need to make changes in one place, not multiple places.

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+                () -> readerService.readFromFilesContents("src/main/resources"
+                        + "/fruitList.txt"));
+        String expectedMessage = "Can't read from the file: "
+                + "src/main/resources/fruitList.csv";
+        String actualMessage = exception.getMessage();
+        assertFalse(actualMessage.contains(expectedMessage));
+    }
+
+    @Test
+    void readFromFilesContents_emptyFile() {
```
okuzan, line 42: i don't see any positive test here, how can we be sure it works at all? :)

src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
```diff
@@ -0,0 +1,31 @@
...
+        assertEquals(CATEGORIES, reportService.generateReport(Map.of()).get(0));
+        assertEquals("apple,3", reportService.generateReport(Map.of("apple", 3)).get(1));
+        assertEquals("banana,5", reportService.generateReport(Map.of("banana", 5)).get(1));
+        assertEquals("orange,7", reportService.generateReport(Map.of("orange", 7)).get(1));
+    }
+
+    @Test
+    void generateReport_nullMap() {
+        Exception exception = assertThrows(NullPointerException.class,
+                () -> reportService.generateReport(null));
```
okuzan, line 26: it's a bit weird to expect and assert NPE, let's change impl in src/main to throw runtime with some message if this happens

src/main/java/core/basesyntax/service/impl/TransactionProcessorServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    private Map<String, Integer> fruitCounts;
+    private final Map<FruitTransaction.Operation, OperationStrategy> strategies;
+
+    public TransactionProcessorServiceImpl(Map<FruitTransaction.Operation,
+            OperationStrategy> strategies) {
+        this.strategies = strategies;
+    }
+
+    @Override
+    public Map<String, Integer> processTransaction(List<FruitTransaction> transactions) {
```
okuzan, line 20:
```suggestion
    public Map<String, Integer> processTransactions(List<FruitTransaction> transactions) {
```
name is confusing, given that it accepts list

src/main/java/core/basesyntax/service/impl/TransactionProcessorServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        FruitTransaction.Operation typeOfOperation = transaction.getType();
+        String fruit = transaction.getFruit();
+        int quantity = transaction.getQuantity();
+
+        OperationStrategy strategy = strategies.get(typeOfOperation);
+        if (strategy != null) {
+            strategy.apply(fruitCounts, fruit, quantity);
+
+        } else {
+            System.out.println("Invalid operation type: " + typeOfOperation);
```
okuzan, line 40: this is not a way to go, printlns should be removed from "production" code, we can't rely on them. If this is not an expected behaviour better to be upfront about it and 'fail fast' than 'eat it' silently

src/test/java/core/basesyntax/service/impl/TransactionProcessorServiceImplTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Test
+    void processTransaction_invalidOperationType() {
+        FruitTransaction.Operation operation = FruitTransaction.Operation.RETURN;
+        String fruit = "apple";
+        int quantity = 10;
+        FruitTransaction transaction = new FruitTransaction(operation, fruit, quantity);
+
+        singleTransaction.processSingleTransaction(transaction);
+        Assertions.assertNull(Storage.getFruits().get(fruit));
```
okuzan, line 59: it's not ok that nothing happens, it will make it hard to debug, if some of operations got ignored from very long list, it's better to have the program fail and fix them one by one, instead of searching for them on your own

src/test/java/core/basesyntax/service/impl/TransactionProcessorServiceImplTest.java
```diff
@@ -0,0 +1,112 @@
...
+        int quantity = 10;
+        FruitTransaction transaction = new FruitTransaction(operation, fruit,
+                quantity);
+        OperationStrategy testStrategy = new OperationStrategy() {
+            @Override
+        public void apply(Map<String, Integer> fruitCountsDifferent,
+                                  String fruit, int quantity) {
+                    fruitCounts.put(transaction.getFruit(), transaction.getQuantity());
+                }
+            };
```
okuzan, line 105: extract it from method, it's hard to read

okuzan (changes requested):

okuzan (changes requested):
strategy handlers are important part of app and should be tested too, write in the chat if you have problems with coverage

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
misterSokor: created one positive method to test if method readFromFilesContents is reading properly

src/test/java/core/basesyntax/service/impl/TransactionProcessorServiceImplTest.java
misterSokor: I decided to delete this test because in the implementation, the processTransaction method in src/main uses the processSingleTransactions method, which handles invalid operation types. Therefore, it seems redundant to me

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class WriterServiceImplTest {
+    private WriterService writerService;
+    private List<String> originalContent;
+
+    private final String filePath = "src/main/resources/output.csv";
```
Elena-Bruyako, line 21: should be constant

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.service.ReaderService;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ReaderServiceImplTest {
+    private final String inputFilePath = "src/main/resources/fruitList.csv";
```
Elena-Bruyako, line 13: should be constants

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/service/impl/WriterServiceImplTest.java
misterSokor: Hi! But this is a constant, isn't?  I know that constants usually use uppercase letters with underscores to separate words but when I do it I get this error:   Member name 'FILE_PATH' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName].  But essentially they are constants. What do you say?

kerrrusha: make it class-level `private static final String FILE_PATH = "src/main/resources/output.csv";`

src/test/java/core/basesyntax/service/impl/ReaderServiceImplTest.java
kerrrusha: same

kerrrusha (changes requested):
suggested constants refactor

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    private static final String CATEGORIES = "fruit,quantity";
+
+    @Override
+    public List<String> generateReport(Map<String, Integer> fruitCounts) {
+        if (fruitCounts == null) {
+            throw new NullPointerException("Can't generate report for null data");
+        } else {
```
okuzan, line 15:
```suggestion
    if (fruitCounts == null) {
        throw new IllegalArgumentException("Can't generate report for null data");
    }
```
We don't deal with NPE explicitly - it happens due to programmer's mistakes, and should be avoided. + We don't need else after throwing/returning

src/test/java/core/basesyntax/service/impl/ReportServiceImplTest.java
okuzan: not fixed, expect illegal argument ex here

src/test/java/core/basesyntax/strategy/SupplyStrategyTest.java
```diff
@@ -0,0 +1,30 @@
...
+    void setUp() {
+        supplyStrategy = new SupplyStrategy();
+        fruitCounts = new HashMap<>();
+        fruitCounts.put("banana", 20);
+        fruitCounts.put("apple", 10);
+
+    }
+
+    @Test
+    void apply() {
```
okuzan, line 24: fix naming for strategies

okuzan (changes requested):

okuzan (approved):

PR #970, opened 2024-04-22
jv-fruti-shop-test

YanaBN:


src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    private FruitTransaction.Operation operation;
+
+    public DataConverterImpl() {
+        this.fruitTransactions = new ArrayList<>();
+    }
+
+    @Override
+    public List<FruitTransaction> convertToTransaction(List<String> transactionData) {
+        if (transactionData.isEmpty()) {
+            throw new RuntimeException("CSV file is empty!" + transactionData);
```
Rommelua, line 26: there is no File here
Delete this IF at all. For empty transactionData return empty List<FruitTransaction> without exception.

src/test/java/core/basesyntax/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    void emptyFile_NotOK() {
+        assertThrows(
+                RuntimeException.class, () -> dataConverter.convertToTransaction(
+                        Collections.emptyList()
+                ));
+    }
+
+    @Test
+    void wrongTextFormat_NotOK() {
```
Rommelua, line 61:
```suggestion
    void convertToTransaction_wrongTextFormat_notOk() {
```

image

Rommelua (changes requested):
Fix all test method names according to the requirements.
image

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
YanaBN: I don't get, why? We can't work with empty data, then if it's empty it should be a misstake, no? The whole point of converting data is so we can read it and commit operations with it. What's the point of returning an empty list then?
Mb I'm mistaking, can you, please, explain this more to me?

Rommelua (changes requested):
Fix all test method names according to the requirements.

Rommelua (approved):

PR #993, opened 2024-06-24
added solution of task

apanyutin:


src/test/java/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    void convertToTransaction_incorrectDataLine_notOk() {
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(
+                List.of("type,fruit,quantity", "b,banana20")));
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(
+                List.of("type,fruit,quantity", "b,banana,20 b,apple,100")));
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(
+                List.of("type,fruit,quantity", "")));
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(
+                List.of("type,fruit,quantity", "b;banana;20")));
+    }
```
MentorMateAcademy, line 36: In one method should use one test case

src/test/java/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+                List.of("type,fruit,quantity", "b,banana,20.00")));
+    }
+
+    @Test
+    void convertToTransaction_Ok() {
+        List<FruitTransaction> expectedList = List.of(
+                new FruitTransaction(Operation.BALANCE, "banana", 20),
+                new FruitTransaction(Operation.SUPPLY, "banana", 100),
+                new FruitTransaction(Operation.PURCHASE, "banana", 15),
+                new FruitTransaction(Operation.RETURN, "apple", 10)
```
MentorMateAcademy, line 60: Save strings to constants

src/test/java/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+                "r,apple,10"
+        );
+        for (int i = 0; i < expectedList.size(); i++) {
+            assertEquals(expectedList.get(i).getOperation(),
+                    dataConverter.convertToTransaction(inputLines).get(i).getOperation());
+            assertEquals(expectedList.get(i).getFruit(),
+                    dataConverter.convertToTransaction(inputLines).get(i).getFruit());
+            assertEquals(expectedList.get(i).getQuantity(),
+                    dataConverter.convertToTransaction(inputLines).get(i).getQuantity());
+
```
MentorMateAcademy, line 76: don't use for, make this logic more simple

src/test/java/service/impl/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,24 @@
...
+    private static ReportGenerator reportGenerator;
+
+    @BeforeAll
+    static void beforeAll() {
+        reportGenerator = new ReportGeneratorImpl();
+    }
+
+    @Test
+    void getReport_consistFirstLine_Ok() {
+        Storage.reports.clear();
```
MentorMateAcademy, line 20: you can make clear beforeEach or afterEach

MentorMateAcademy (changes requested):

src/test/java/service/impl/DataConverterImplTest.java
apanyutin: Why? I guess in this case will be better put all this incorrect strings into one method.

src/test/java/service/impl/DataConverterImplTest.java
apanyutin: How can I compare these lists? I can compare expected List with result List when it have overridden method equals, but  it don't have. How to simple this code?

src/test/java/service/impl/fromTestFile.csv
```diff
@@ -0,0 +1,10 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
+p,banana,12
```
sarakhmen, line 10: this file shouldn't be pushed to the remote repo :)

src/test/java/service/impl/fromWrongTestFile.csv
```diff
@@ -0,0 +1,9 @@
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
```
sarakhmen, line 5: and this as well. Please fix such issues for all cases

src/test/java/strategy/impl/ReturnOperationTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import db.Storage;
+import java.util.Map;
+import model.FruitTransaction;
+import model.Operation;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ReturnOperationTest {
```
sarakhmen, line 13: please clear your storage after each test method

sarakhmen: fix same issue in all places

sarakhmen (changes requested):
let's improve your solution :)

src/test/java/service/impl/DataConverterImplTest.java
sarakhmen: in general, it's better to divide your test logic into several methods with proper names. But in this case I think it's enough to leave everything as is

src/test/java/service/impl/DataConverterImplTest.java
sarakhmen: but you're duplicating conversion logic a lot of times

src/test/java/service/impl/fromTestFile.csv
apanyutin: What does it mean? Do I need to create a folder for these files?

apanyutin: How can it test FileReader and FileWriter without these files?

sarakhmen: sorry for unclear explanation :) Yes, you should create a separate resources folder and place it there.

sarakhmen (approved):

PR #995, opened 2024-06-27
First attempt, Tests added

jszymanski97:


MateuszRostek (changes requested):
Do not create PR before making sure all checks have passed.

jszymanski97:
> Do not create PR before making sure all checks have passed.

They are passing for me, I was surprised they didn't pass here hence my post on the chat about this issue

src/main/java/core/basesyntax/service/impl/TransitionProcessorImpl.java
```diff
@@ -0,0 +1,26 @@
...
+                        balanceMap.getOrDefault(fruit, 0) + quantity);
+            case PURCHASE -> balanceMap.put(fruit,
+                        balanceMap.getOrDefault(fruit, 0) - quantity);
+            default -> throw new IllegalArgumentException("Invalid transaction type: "
+                        + transaction.getOperation());
+        }
+
+        if (balanceMap.get(fruit) < 0) {
+            balanceMap.put(fruit, 0);
+        }
```
MateuszRostek, line 24: Just adding 0 when there is an error in processing transaction is not a good approach, you should throw an exception for example.

src/test/java/core/basesyntax/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,67 @@
...
+        List<FruitTransaction> actual = converter.convert(correctList);
+
+        for (int i = 0; i < expectedFruitTransactionList.size(); i++) {
+            assertEquals(expectedFruitTransactionList.get(i).getFruit(),
+                    actual.get(i).getFruit());
+            assertEquals(expectedFruitTransactionList.get(i).getOperation(),
+                    actual.get(i).getOperation());
+            assertEquals(expectedFruitTransactionList.get(i).getQuantity(),
+                    actual.get(i).getQuantity());
+        }
```
MateuszRostek, line 38: Look for a better way to compare 2 lists. Comparing them one by one in a for loop like this is not a good approach.

src/test/java/core/basesyntax/service/impl/DataProcessImplTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        Map<String, Integer> expectedData = new HashMap<>();
+        expectedData.put("banana", 20);
+
+        Map<String, Integer> actual = dataProcess.process(transactionList);
+        assertEquals(expectedData, actual);
+    }
+
+    @Test
+    void process_minusQuantityTransaction_notOk() {
```
MateuszRostek, line 61: As I said, just setting the value to 0 in a case with negative quantity is not a good approach.

MateuszRostek (changes requested):
Nice try, but there are some changes you should make:

Where are tests for `TransitionProcessorImpl`? Also you should check and test what happens when the passed value is negative OR you get negative result after the PURCHASE operation.

Also, why you didn't test `ReportToFileImpl` class? It should also be tested.

src/test/java/core/basesyntax/service/impl/ReportToFileImplTest.java
MateuszRostek, line 1: Why do you have only one tests here?
You should consider creating more tests, for example:
`..._nullInput_notOk`
`..._wrongFileExtension_notOk`
`..._ioExceptionThrown_notOk()`
`..._validData_ok`

Also, you shouldn't have any try-catch blocks inside tests.

MateuszRostek (approved):
Good job! There is one thing you should change in the `ReportToFileImplTest`, but I'm sure you can handle it, so approved!

PR #1011, opened 2024-08-14
Result

ADIGrimm:


src/test/resources/InputFile.csv
```diff
@@ -0,0 +1,7 @@
+header
```
okuzan, line 1: name of the file should resemble that it's corrupted file

okuzan (approved):
good job overall!

PR #1017, opened 2024-08-26
Home work fruit shop tests

4Vitalii5:
I created the tests to cover almost all the classes from my previous program.

### DIFF
--- a/src/test/java/dao/FruitDaoImplTest.java
+++ b/src/test/java/dao/FruitDaoImplTest.java
@@ -1,0 +1,96 @@
+package dao;
+
+import db.Storage;
+import java.util.AbstractMap;
+import java.util.Map;
+import java.util.Set;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class FruitDaoImplTest {
+    private static FruitDao fruitDao;
+
+    @BeforeAll
+    static void beforeAll() {
+        fruitDao = new FruitDaoImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        Storage.fruitStock.put("apple", 120);
+        Storage.fruitStock.put("banana", 65);
+    }
+
+    @Test
+    void getBalance_validFruitName_isOk() {
+        String expectedName = "apple";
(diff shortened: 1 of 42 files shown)


src/test/java/dao/FruitDaoImplTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @BeforeAll
+    static void beforeAll() {
+        fruitDao = new FruitDaoImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        Storage.fruitStock.put("apple", 120);
+        Storage.fruitStock.put("banana", 65);
```
MentorMateAcademy, line 24: Save strings to constants

src/test/java/dao/FruitDaoImplTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    void getBalance_validFruitName_isOk() {
+        String expectedName = "apple";
+        int expectedQuantity = 120;
+        String actualName = Storage.fruitStock.keySet().stream()
+                .filter(k -> k.equals(expectedName))
+                .findFirst()
+                .get();
```
MentorMateAcademy, line 34: you just can get from map

src/test/java/service/impl/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+            public void updateBalance(String fruit, int quantity) {
+
+            }
+
+            @Override
+            public Set<Map.Entry<String, Integer>> getAllEntries() {
+                return Storage.fruitStock.entrySet();
+            }
+        };
+        reportGenerator = new ReportGeneratorImpl(fruitDao);
```
MentorMateAcademy, line 41: why do you implement annonimous class?

MentorMateAcademy (changes requested):

src/test/java/dao/FruitDaoImplTest.java
4Vitalii5: If you can, please show me because I didn't find the way to do it.

src/test/java/service/impl/ReportGeneratorImplTest.java
4Vitalii5: I used them to make my test class independent from the FruitDaoImpl. Is this wrong?

okuzan (approved):

PR #1023, opened 2024-09-04
Added task solution

Lessorg:


src/test/java/core/basesyntax/service/impl/ShopServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+        RuntimeException exception = assertThrows(RuntimeException.class, ()
+                -> shopService.proceedAll(transactions));
+
+        assertTrue(exception.getMessage().contains("Transaction is null"));
+    }
+
+    private static class InMemoryFruitDao implements FruitDao {
+        private final Map<String, Integer> fruitStorage = new HashMap<>();
+
+        @Override
```
okuzan, line 74: why do you provider you custom impelemntation here? try to simplify the test. Use existing implementation if you test them, mock everything else

src/test/java/core/basesyntax/model/FruitTransactionTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class FruitTransactionTest {
```
okuzan, line 11: we don't test models, they are considered to be tested indirectly via services that use them

okuzan (changes requested):

PR #1024, opened 2024-09-04
Add JUnit test coverage for the fruit shop project

smagles:


MateuszRostek (approved):
Good job! Consider this:

What about test cases in operations when quantity is < 0? You should add tests for it and check if your code correctly handles the negative amount (some exception should be thrown).

PR #1025, opened 2024-09-05
Created test classes

svitlanagalicka:


okuzan (changes requested):
build failed, you can check it locally with `mvn clean package`

PR #1064, opened 2024-11-07
first version

MateuszSeler:


slade13 (approved):
It would be good to also include Report service in your tests, but in overall your tests are looking good and code coverage for testing services is above 80% so it's fine! You can still think about adding tests for Report service if you feel strong enough, you can try to do that. 👍

PR #1069, opened 2024-11-11
Jv fruit shop tests

LynAlinka:


okuzan (changes requested):
Hi, CI checks failed, please make sure the project builds locally without issues before re-requesting reviews, with `mvn clean package` command

LynAlinka:
@okuzan I can't figure out what the error is when I write the command mvn clean package there is build successful. But still no coverage of test classes

ahoienko (changes requested):
Build still fails, please fix build first

PR #1081, opened 2024-11-28
resolvee

DankevichMisha:


src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,64 @@
...
+                FruitTransaction.Operation.BALANCE, "banana", -100);
+        assertThrows(InvalidDataException.class,
+                () -> balanceOperationHandler.calculateOperation(bananaTransaction));
+    }
+
+    @Test
+    void addFruits_nullFruit_notOk() {
+        FruitTransaction bananaTransaction = new FruitTransaction(
+                FruitTransaction.Operation.BALANCE, null, 100);
+        assertThrows(NullPointerException.class,
```
Elena-Bruyako, line 53: it's bad practice to expect NullPointerException, it should be covered in main code to avoid this exception

src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,64 @@
...
+                () -> balanceOperationHandler.calculateOperation(bananaTransaction));
+    }
+
+    @Test
+    void addFruit_nullOperationType_notOk() {
+        FruitTransaction bananaTransaction = new FruitTransaction(
+                null, "banana", 100);
+        assertThrows(NullPointerException.class,
```
Elena-Bruyako, line 61: same here

Elena-Bruyako (changes requested):
looks good, let's fix comments about NullPointerException

src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,48 @@
...
+        FruitTransaction bananaTransaction = new FruitTransaction(
+                FruitTransaction.Operation.BALANCE, "banana", 120);
+        assertDoesNotThrow(() -> balanceOperationHandler.calculateOperation(bananaTransaction));
+        FruitTransaction appleTransaction = new FruitTransaction(
+                FruitTransaction.Operation.BALANCE, "apple", 50);
+        assertDoesNotThrow(() -> balanceOperationHandler.calculateOperation(appleTransaction));
+        int expectedBanana = Storage.STORAGE.get("banana");
+        assertEquals(expectedBanana, 120);
+        int expectedApple = Storage.STORAGE.get("apple");
+        assertEquals(expectedApple, 50);
```
ahoienko, line 38: Please add some line break to make this code look better

src/test/java/core/basesyntax/services/DataProcessorServiceTest.java
```diff
@@ -0,0 +1,55 @@
...
+    void processData_invalidQuantity_notOk() {
+        List<String> data = List.of("b,banana,-200", "b,apple,20", "s,banana,10");
+        assertThrows(RuntimeException.class, () -> dataProcessorService.processInputData(data));
+    }
+
+    @Test
+    void processData_invalidOperationType_notOk() {
+        List<String> data = List.of("buy, banana,-200", "b,apple,20", "s,banana,10");
+        assertThrows(RuntimeException.class, () -> dataProcessorService.processInputData(data));
+    }
```
ahoienko, line 48: there is no reason to handle some general error, if you want this case to be handled by your code, please create you own exception throw it with some error message and check in tests

src/test/java/core/basesyntax/services/ReportCreatorServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+        assertEquals(expected, reportCreatorService.createReport());
+    }
+
+    @Test
+    void creatorReport_existingData_ok() {
+        Storage.STORAGE.put("banana", 10);
+        Storage.STORAGE.put("apple", 20);
+        String expectedReport = "fruit,quantity" + System.lineSeparator()
+                + "banana,10" + System.lineSeparator() + "apple,20" + System.lineSeparator();
+        assertEquals(expectedReport, reportCreatorService.createReport());
```
ahoienko, line 36:
```suggestion
        String actualReport  = reportCreatorService.createReport()
        assertEquals(expectedReport, actualReport  );
```

ahoienko: This is much readable and clear

ahoienko (changes requested):
Nice, left some comments!

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> STORAGE = new HashMap<>();
```
JJJazl, line 7: It is not a constant because constant means it does not change, but you can add or remove values ​​from the map, so it is modifiable and can be a constant.
Check Collection.emptyMap() or Map.of() methods. They return an immutable map
```suggestion
    public static final Map<String, Integer> storage = new HashMap<>();
```

src/main/java/core/basesyntax/handlers/ReturnOperationHandler.java
```diff
@@ -0,0 +1,21 @@
...
+import core.basesyntax.strategy.OperationHandler;
+
+public class ReturnOperationHandler implements OperationHandler {
+    @Override
+    public void calculateOperation(FruitTransaction transaction) {
+        if (transaction.getQuantity() < 0) {
+            throw new InvalidDataException("Negative quantity");
+        } else if (transaction.getFruit() == null) {
+            throw new IllegalArgumentException("Invalid fruit type");
+        }
```
JJJazl, line 15: I guess it will be better. The same in other handlers
```suggestion
        if (transaction.getQuantity() < 0) {
            throw new InvalidDataException("Negative quantity");
        } 
        if (transaction.getFruit() == null) {
            throw new IllegalArgumentException("Invalid fruit type");
        }
```

src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    void addFruits_NegativeQuantity_notOk() {
+        FruitTransaction bananaTransaction = new FruitTransaction(
+                FruitTransaction.Operation.BALANCE, "banana", -100);
+        assertThrows(InvalidDataException.class,
```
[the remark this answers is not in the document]
JJJazl: Not fixed. Please don't resolve the conversation yourself, just write a comment or mark it with an emoji

src/test/java/core/basesyntax/handlers/BalanceOperationHandlerTest.java
```diff
@@ -0,0 +1,51 @@
...
+        );
+        assertDoesNotThrow(() -> balanceOperationHandler.calculateOperation(bananaTransaction));
+        FruitTransaction appleTransaction = new FruitTransaction(
+                FruitTransaction.Operation.BALANCE, "apple", 50
+        );
+        assertDoesNotThrow(() -> balanceOperationHandler.calculateOperation(appleTransaction));
+        int expectedBanana = Storage.STORAGE.get("banana");
+        assertEquals(120, expectedBanana);
+        int expectedApple = Storage.STORAGE.get("apple");
+        assertEquals(50, expectedApple);
```
JJJazl, line 40: We don't need to use `assertDoesNotThrow()` in this case, because we expect that exception will not thrown by default and if we expect that exception will be thrown we use `assertThrows()`. Please check everywhere
```suggestion
        FruitTransaction bananaTransaction = new FruitTransaction(
                FruitTransaction.Operation.BALANCE, "banana", 120
        );
        FruitTransaction appleTransaction = new FruitTransaction(
                FruitTransaction.Operation.BALANCE, "apple", 50
        );
   
        balanceOperationHandler.calculateOperation(bananaTransaction)
        balanceOperationHandler.calculateOperation(appleTransaction)
        
        int expectedBanana = Storage.STORAGE.get("banana");
        int expectedApple = Storage.STORAGE.get("apple");
        assertEquals(120, expectedBanana);
        assertEquals(50, expectedApple);
```

src/test/java/core/basesyntax/services/DataProcessorServiceTest.java
JJJazl: Not fixed

JJJazl (changes requested):

src/test/java/core/basesyntax/services/DataProcessorServiceTest.java
DankevichMisha: I try this and get such error 
DataProcessorServiceTest.processData_invalidOperationType_notOk:48 Unexpected exception type thrown, expected: <core.basesyntax.exception.InvalidOperationTypeException> but was: <java.lang.RuntimeException>

Elena-Bruyako (approved):

PR #1100, opened 2025-01-12
test for shop service

katrienkraska:


liliia-ponomarenko (changes requested):
Firstly, let’s fix the build!

PR #1110, opened 2025-01-25
add shop and tests

FroGitHub:


src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,7 +1,206 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Test;
+
 public class HelloWorldTest {
```
liliia-ponomarenko, line 25: Your tests should have the same structure as that main solution. Create separate classes.

liliia-ponomarenko (changes requested):
Firstly, let’s fix the test structure

Elena-Bruyako (changes requested):
add tests for other services

Elena-Bruyako (approved):

PR #1128, opened 2025-03-06
Created tests for jv-fruit-shop, united in FinalTest to be able to te…

serhii-hl:
…st the code in one class.

### DIFF
--- a/src/main/java/core/basesyntax/dao/CsvWriterImpl.java
+++ b/src/main/java/core/basesyntax/dao/CsvWriterImpl.java
@@ -19,5 +19,3 @@ public class CsvWriterImpl implements CustomFileWriter {
         }
     }
 }
-
-


src/test/java/core/basesyntax/service/CsvReaderImplTest.java
```diff
@@ -0,0 +1,34 @@
...
+        assertThrows(RuntimeException.class, () -> {
+            csvReader.readFile(filePath);
+        });
+    }
+
+    @Test
+    void file_checkOutput_IsOk() throws IOException {
+        String expected = "type,fruit,quantity" + System.lineSeparator()
+                + "b,banana,20" + System.lineSeparator();
+        System.out.println(csvReader.readFile("src/test/java/testResources/testFile.csv"));
```
Elena-Bruyako, line 31: remove System.out.println
check all places

src/test/java/core/basesyntax/service/CsvWriterImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+        String filePath = null;
+        List<String> list = new ArrayList<>();
+        assertThrows(RuntimeException.class, () -> {
+            csvWriter.writeFile(filePath, list);
+        });
+    }
+
+    @Test
+    void file_checkOutput_IsOk() throws IOException {
+        String filePath = "src/test/java/testResources/testWrittenFile.csv";
```
Elena-Bruyako, line 35: create constants for file pathes

src/test/java/core/basesyntax/service/FinalTest.java
```diff
@@ -0,0 +1,45 @@
+package core.basesyntax.service;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+public class FinalTest {
```
Elena-Bruyako, line 6: what is this test about?

src/test/java/core/basesyntax/service/ReportGeneratorTest.java
```diff
@@ -0,0 +1,27 @@
...
+        Storage.fruitsStorage.put("banana", 20);
+        Map<String, Integer> map = Storage.fruitsStorage;
+        ReportCreator reportCreator = new ReportGeneratorImpl();
+        List<String> expected = new ArrayList<>();
+        expected.add("fruit,quantity");
+        expected.add("banana,20");
+        expected.add("apple,10");
+        List<String> actual = reportCreator.createReport(map);
+        assertEquals(expected, actual);
+        Storage.fruitsStorage.clear();
```
Elena-Bruyako, line 25: clear should be in @AfterEach method

src/test/java/core/basesyntax/service/ReportGeneratorTest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class ReportGeneratorTest {
+    @Test
+    void report_Generating_IsOk() {
+        Storage.fruitsStorage.put("apple", 10);
+        Storage.fruitsStorage.put("banana", 20);
+        Map<String, Integer> map = Storage.fruitsStorage;
+        ReportCreator reportCreator = new ReportGeneratorImpl();
```
Elena-Bruyako, line 18: replace this set up to @BeforeAll method

src/test/java/core/basesyntax/service/FileFormaterTest.java
```diff
@@ -0,0 +1,34 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class FileFormaterTest {
+    @Test
+    void file_Parsing_IsOk() {
+        CustomFileReader reader = new CsvReaderImpl();
+        FileFormaterForCsvReader fileFormaterForCsvReader = new FileFormaterForCsvReader(reader);
```
Elena-Bruyako, line 20: it should be in BeforeAll method
check all test classes

Elena-Bruyako (changes requested):
add tests for each handler and service

src/test/java/core/basesyntax/service/FinalTest.java
serhii-hl: I added this test to run all tests in one click, should we run every test separately or is there any other way to unite all tests? Thanks!

serhii-hl: I removed it, so please let me know if we need this in tests or run every test separately is okay, thanks for your advices!

src/test/java/core/basesyntax/service/CsvWriterImplTest.java
```diff
@@ -0,0 +1,48 @@
+package core.basesyntax.service;
```
liliia-ponomarenko, line 1: your test clases should have the same structer that main solution

src/test/java/core/basesyntax/service/BalanceCalculatorTest.java
```diff
@@ -0,0 +1,97 @@
...
+        String fruit = "apple";
+        OperationsList operation = OperationsList.PURCHASE;
+        int quantity = 10;
+        FruitTransaction fruitTransaction = new FruitTransaction();
+        fruitTransaction.setOperation(operation);
+        fruitTransaction.setFruit(fruit);
+        fruitTransaction.setQuantity(quantity);
+        List<FruitTransaction> list = new ArrayList<>();
+        list.add(fruitTransaction);
+        assertEquals(-10, balanceCalculator.update(list).get(fruit));
```
liliia-ponomarenko, line 49: Is it normal that fruit can have a negative value on the balance?

src/test/java/core/basesyntax/service/BalanceCalculatorTest.java
```diff
@@ -0,0 +1,97 @@
...
+import core.basesyntax.service.impl.ReturnHandler;
+import core.basesyntax.service.impl.SupplyHandler;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class BalanceCalculatorTest {
```
liliia-ponomarenko, line 20: Add a test case when the storage already has some quantity of fruits

src/test/java/core/basesyntax/service/BalanceHandlerTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.service.impl.BalanceHandler;
+import org.junit.jupiter.api.Test;
+
+public class BalanceHandlerTest {
+    @Test
+    void balance_Check_IsOk() {
+        BalanceHandler balanceHandler = new BalanceHandler();
```
liliia-ponomarenko, line 11: Make it a class-level variable

src/test/java/core/basesyntax/service/CsvReaderImplTest.java
```diff
@@ -0,0 +1,34 @@
...
+    private CsvReaderImpl csvReader;
+
+    @BeforeEach
+    void setUp() {
+        csvReader = new CsvReaderImpl();
+    }
+
+    @Test
+    void file_isNull_NotOk() {
+        String filePath = null;
```
liliia-ponomarenko, line 22: Do you have such a check in your solution? Don’t  test system exceptions

src/test/java/core/basesyntax/service/CsvReaderImplTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.dao.CsvReaderImpl;
+import java.io.IOException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CsvReaderImplTest {
```
liliia-ponomarenko, line 11: add cases with wrong path and empty file. You should test all edge cases.

src/test/java/core/basesyntax/service/CsvWriterImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CsvWriterImplTest {
```
liliia-ponomarenko, line 16: the same comments

src/test/java/core/basesyntax/service/CsvWriterImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    void file_checkOutput_IsOk() throws IOException {
+        List<String> list = new ArrayList<>();
+        list.add("banana,20");
+        list.add("apple,40");
+        File file = new File(ACTUAL_PATH);
+        csvWriter.writeFile(ACTUAL_PATH, list);
+        List<String> expected = Files.readAllLines(
```
liliia-ponomarenko, line 42: Don’t read the expected content from the file. initialize this value directly

src/test/java/core/basesyntax/service/FileFormaterTest.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Test
+    void file_Parsing_IsOk() {
+        try (BufferedWriter writer = Files.newBufferedWriter(Paths.get(FILE_PATH))) {
+            writer.write("operation,fruit,quantity\n");
+            writer.write("s,apple,20\n");
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
liliia-ponomarenko, line 35: Don’t complicate your test cases. Provide file with content

src/test/java/core/basesyntax/service/PurchaseHandlerTest.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.service.impl.PurchaseHandler;
+import org.junit.jupiter.api.Test;
+
+public class PurchaseHandlerTest {
+    @Test
+    void return_Check_IsOk() {
```
liliia-ponomarenko, line 10: add negative test case

src/test/java/core/basesyntax/service/ReportGeneratorTest.java
```diff
@@ -0,0 +1,40 @@
...
+    @BeforeEach
+    void setUp() {
+        Storage.fruitsStorage.put("apple", 10);
+        Storage.fruitsStorage.put("banana", 20);
+        map = Storage.fruitsStorage;
+        reportCreator = new ReportGeneratorImpl();
+    }
+
+    @Test
+    void report_Generating_IsOk() {
```
liliia-ponomarenko, line 27: add case for empty storage

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution ;)

src/test/java/core/basesyntax/service/BalanceCalculatorTest.java
serhii-hl: yes, because we can have the first operation purchase and the second as supply, but at the end we should have a debit balance, which I should check

src/test/java/core/basesyntax/service/BalanceCalculatorTest.java
serhii-hl: As I understand, storage should be empty before we do balancecalculations, so we can have storage filled from our map, that is why can I edit the original code and add the check if the storage is empty and if it is not, clear it? Thanks!

src/test/java/core/basesyntax/service/CsvReaderImplTest.java
serhii-hl: yes, I have a check if the file path is not found, I just unsure what name I should use, so I used null. Let me replace for some random name

src/test/java/core/basesyntax/service/FinalTest.java
Elena-Bruyako: No, you don't need to have this test

Elena-Bruyako (changes requested):

serhii-hl:
Thanks, I renamed the methods and moved initialisations to beforeAll

src/test/java/core/basesyntax/service/BalanceCalculatorTest.java
liliia-ponomarenko: It depends on how you interpret it. In my opinion, the balance is a certain starting point. That is, you could have 10 bananas, but 3 had to be written off, so the balance transaction will contain 7 and you have to set this value, regardless of how many fruits there were before. But you can interpret it differently. Clear requirements are not specified in the task.

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

serhii-hl:
You are right, I was thinking that it would help if balance can be after purchase but I didn`t think what if at the end of the end the balance will be negative. I modified my code, thanks for your help, Liliia!

src/main/java/core/basesyntax/service/impl/PurchaseHandler.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.OperationHandler;
+
+public class PurchaseHandler implements OperationHandler {
+    @Override
+    public int operate(int operationAmount, int balanceAmount) {
+        if (operationAmount < 0) {
```
liliia-ponomarenko, line 8:
```suggestion
        if (operationAmount < 0 || balanceAmount < operationAmount) {
```

src/main/java/core/basesyntax/service/impl/PurchaseHandler.java
```diff
@@ -0,0 +1,13 @@
...
+            throw new RuntimeException("Balance amount is less than 0, please check your file");
```
liliia-ponomarenko, line 9:
```suggestion
            throw new RuntimeException("The purchase amount is less than 0 or less than the current balance. Balance: " + balanceAmount + " operation amount: " + operationAmount);
```

src/main/java/core/basesyntax/service/impl/BalanceChecker.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.impl;
+
+public class BalanceChecker {
+    void checkBalance(int balance, String fruit) {
+        if (balance < 0) {
+            throw new RuntimeException("Balance is negative, please check your file for fruit: "
+            + fruit);
+        }
```
liliia-ponomarenko, line 8: This check does not guarantee that the data in the file is valid. At some point in time, the balance could have been negative, but you would not know this because you are performing the check after all transactions have been applied.

src/main/java/core/basesyntax/service/impl/BalanceChecker.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.impl;
+
+public class BalanceChecker {
```
liliia-ponomarenko, line 3: See comments to PurchaseHandler

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

src/main/java/core/basesyntax/service/impl/BalanceChecker.java
serhii-hl: Got it, my check is about the total balance after all operations, but we can have a situation when balance is -10, and return is 20, so it will show us 10 which is correct for check, but balance can`t be negative. Let me fix this as well, thanks for explaining!

serhii-hl: I checked my BalanceHandlerTest, it checks if the balance is negative when run the stratgey, so if balance is -10, it will throw an exception:     @Test
    void operate_OperatingAmountNegative_NotOk() {
        assertThrows(RuntimeException.class, () -> {
            balanceHandler.operate(-10, 20);
        });
    }
Balancechecker is working on checking if the total more than 0, so in case of balance 10, purchase 20 it will throw an exception

src/main/java/core/basesyntax/service/impl/PurchaseHandler.java
serhii-hl: Sorry, I am not sure if this should be checked, as I understand, initial file should contain all transactions from the shop, so they can have a balance 20, purchase 30 and return 10 ( it can have 3rd position in the list, but in fact it could be even first in the day, but the salesman added it on the thrid position ) which makes 0, or is it works like a stack, so we should check operation one by one, like we can`t sell 30 from 20?

Elena-Bruyako (approved):

PR #1176, opened 2025-06-22
executed2222

Yevhen4444:


liliia-ponomarenko (changes requested):
Firstly, let's fix the build

PR #1201, opened 2025-10-09
implemented tests

chernykh-mykhailo:


src/test/java/core/basesyntax/model/FruitTransactionTest.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class FruitTransactionTest {
```
sarakhmen: no need to create a test for this class. It doesn't contain any useful/business logic

src/test/java/core/basesyntax/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    void convertToTransaction_ValidData_ReturnsCorrectTransactions() {
+        List<String> input = List.of(
+                "type,fruit,quantity",
+                "b,banana,20",
+                "p,apple,10"
+        );
+        List<FruitTransaction> transactions = dataConverter.convertToTransaction(input);
+        assertEquals(2, transactions.size());
+        assertEquals(FruitTransaction.Operation.BALANCE, transactions.get(0).getOperation());
```
sarakhmen: you should compare the entire dataset. If the result is a list - you should compare the whole list

src/test/java/core/basesyntax/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void convertToTransaction_InvalidHeader_ThrowsRuntimeException() {
+        List<String> input = List.of(
+                "invalid,header",
+                "b,banana,20"
+        );
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(input));
```
sarakhmen: As you can see the exception that is thrown is not informative enough. So, in this case it's better also compare the error message or/and create a separate exception class and throw it in a logic

src/test/java/core/basesyntax/service/impl/DataConverterImplTest.java
sarakhmen: fix same issue in all places

src/test/java/core/basesyntax/service/impl/FileReaderImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+        Files.write(testPath, expectedLines);
+
+        List<String> actualLines = fileReader.read(TEST_FILE_PATH);
+        assertEquals(expectedLines, actualLines, "Content read should match content written.");
+    }
+
+    @Test
+    void read_NonExistentFile_ThrowsRuntimeException() throws IOException {
+        Files.deleteIfExists(testPath);
+
```
sarakhmen, line 46: do you really need it here? it's already present in teardown method

src/test/java/core/basesyntax/service/impl/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @BeforeEach
+    void setUp() {
+        reportGenerator = new ReportGeneratorImpl();
+        Storage.setFruitQuantity("apple", 30);
+        Storage.setFruitQuantity("banana", 50);
+    }
+
+    @AfterEach
+    void tearDown() {
+    }
```
sarakhmen: remove this empty teardown method

sarakhmen: or consider adding some useful logic inside of it. I think it'd be great to clear storage here

src/test/java/core/basesyntax/service/impl/ShopServiceImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+        // ... додайте всі інші обробники ...
+
+        // 2. Створюємо справжню стратегію
+        OperationStrategy operationStrategy = new OperationStrategyImpl(handlers);
+
+        // 3. Створюємо справжній сервіс
+        shopService = new ShopServiceImpl(operationStrategy);
+
+        // Оскільки ми не можемо очистити сховище, ми встановлюємо початкове значення 0
+        Storage.setFruitQuantity(APPLE, 0);
```
sarakhmen: you're using chatgpt or similar llm for your HW :(((((( let it remain our secret, but don't abuse it anymore.
Please init your shopService in a separate setUp method, as in other tests

sarakhmen (changes requested):
Let's improve your solution :)

KuryloVolodymyr (approved):
good job

PR #1203, opened 2025-10-11
Add comprehensive unit tests to the project

MaksymMaksymenko2:
Add comprehensive unit tests to the project


src/test/java/db/StorageTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class StorageTest {
```
sarakhmen: no need to test this class, we can remove this test

src/test/java/model/FruitTransactionTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+public class FruitTransactionTest {
+
+    @Test
+    void constructor_validData_fieldsAreSetCorrectly() {
+        FruitTransaction transaction = new FruitTransaction(
```
sarakhmen: this test class - as well

src/test/java/service/impl/DataConverterImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        FruitTransaction second = result.get(1);
+        assertEquals(FruitTransaction.Operation.SUPPLY, second.getOperation());
+        assertEquals("apple", second.getFruit());
+        assertEquals(15, second.getQuantity());
+
+        FruitTransaction third = result.get(2);
+        assertEquals(FruitTransaction.Operation.PURCHASE, third.getOperation());
+        assertEquals("banana", third.getFruit());
+        assertEquals(5, third.getQuantity());
```
sarakhmen: I'd recomment create a list of expected transactions and comparing them in 1 line (expected with actual result). Creating transactions could be moved to the separate method, utility method or class. Fix in all places

src/test/java/service/impl/ReaderServiceImplTest.java
```diff
@@ -0,0 +1,45 @@
...
+    private final ReaderService readerService = new ReaderServiceImpl();
+
+    @Test
+    void readFromFile_validFile_returnsAllLines() throws IOException {
+        Path tempFile = Files.createTempFile("test-file", ".csv");
+        try (FileWriter writer = new FileWriter(tempFile.toFile())) {
+            writer.write("type,fruit,quantity\n");
+            writer.write("b,banana,20\n");
+            writer.write("s,apple,15\n");
+        }
```
sarakhmen: let's use reference to the test resources instead (src/test/resource). Use system.lineSeparator() instead of '\n'. Fix similar issues in all places

sarakhmen: Also you can create a file with data for test manually and add it to git

src/test/java/service/impl/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,60 @@
...
+        map.clear();
+    }
+
+    private Map<String, Integer> getFruitStorage() throws Exception {
+        var field = Storage.class.getDeclaredField("fruitStorage");
+        field.setAccessible(true);
+        @SuppressWarnings("unchecked")
+        Map<String, Integer> map = (Map<String, Integer>) field.get(null);
+        return map;
+    }
```
sarakhmen: you can add an aditional public method to the fruitStorage or the field public

sarakhmen: fix similar issue in all places

sarakhmen: also it's better to clear data in aftereach method

sarakhmen: fix similar issues in all places

src/test/java/service/impl/ShopServiceImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+    static class PurchaseHandler implements OperationHandler {
+        @Override
+        public void apply(FruitTransaction transaction) {
+            Storage.addQuantity(transaction.getFruit(), -transaction.getQuantity());
+        }
+    }
+
+    static class ReturnHandler implements OperationHandler {
+        @Override
+        public void apply(FruitTransaction transaction) {
```
sarakhmen: no need to create additional static classes. In the future you will use Mockito to mock such dependencies. For now it's enough to use existing ones

sarakhmen (changes requested):
good job! Let's slighlty improve your solution :)

src/test/java/service/impl/ShopServiceImplTest.java
MaksymMaksymenko2: Thank you. I'll try to fix it.

Elena-Bruyako (approved):

PR #1221, opened 2025-11-05
First solution.

AliakseiS97:


slade13 (changes requested):
Please don't make PR's from your `main` branch. Create a new feature branch then create PR from 'your-feature-branch' to 'forked-repo:main'. 

Please also fix paths in your tests, test classes should be in the same package as the tested subject:
`src/test/java/core/basesyntax/test/service/strategy/BalanceHandlerTest.java`
is testing:
`src/main/java/core/basesyntax/service/strategy/BalanceHandler.java`

so your test should be in package:
`src/test/java/core/basesyntax/service/strategy/BalanceHandlerTest.java`

Please fix it and create a new PR.

PR #1223, opened 2025-11-06
add tests for shop fruit

bim121:


src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+
+    private static final Map<String, Integer> fruitStorage = new HashMap<>();
+
+    private Storage() {
+    }
```
KuryloVolodymyr: empty constructor is generated by default. please remove it

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,62 @@
+package core.basesyntax.model;
+
+public class FruitTransaction {
+    private Operation operation;
+    private String fruit;
+    private int quantity;
+
+    public FruitTransaction() {
+    }
```
KuryloVolodymyr: same as above

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,47 @@
...
+            String line = lines.get(i).trim();
+            if (line.isEmpty()) {
+                continue;
+            }
+            String[] parts = line.split(",");
+            if (parts.length != 3) {
+                throw new IllegalArgumentException("Malformed line at "
+                        + (i + 1) + ": " + line);
+            }
+            String type = parts[0].trim();
```
KuryloVolodymyr, line 23: add validation if a type is empty

src/test/java/core/basesyntax/db/StorageTest.java
```diff
@@ -0,0 +1,39 @@
...
+    @Test
+    void getQuantity_notExists_returnsZero() {
+        assertEquals(0, Storage.getQuantity("orange"));
+    }
+
+    @Test
+    void clear_removesAllEntries() {
+        Storage.setQuantity("banana", 10);
+        Storage.clear();
+        assertTrue(Storage.getAll().isEmpty());
```
KuryloVolodymyr: you should assert if the value exists in storage, before checking if its cleared

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,62 @@
...
+
+        public static Operation fromCode(String code) {
+            for (Operation op : values()) {
+                if (op.code.equals(code)) {
+                    return op;
+                }
+            }
+            throw new IllegalArgumentException("Unknown operation: " + code);
+        }
+    }
```
KuryloVolodymyr: since this class is used in multiple other classes, it makes sense to move it to a separate file. Inner clases make sence if they're used mostly within the cless they're declared in

KuryloVolodymyr (changes requested):
please fix comments

src/test/java/core/basesyntax/db/StorageTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class StorageTest {
+    @BeforeEach
+    void setUp() {
+        Storage.clear();
+    }
```
sarakhmen: let's make your solution perfect. Perfect solution shouldn't contain unit tests for storage and fruittransaction. Instead, you should have tests for ALL services and handlers

src/test/java/core/basesyntax/service/DataConverterImplTest.java
```diff
@@ -0,0 +1,45 @@
...
+        List<String> lines = Arrays.asList(
+                "type,fruit,quantity",
+                "b,banana,20",
+                "s,apple,10"
+        );
+        List<FruitTransaction> result = converter.convertToTransaction(lines);
+        assertEquals(2, result.size());
+        assertEquals(Operation.BALANCE, result.get(0).getOperation());
+        assertEquals("banana", result.get(0).getFruit());
+        assertEquals(20, result.get(0).getQuantity());
```
sarakhmen: I and my colleagues are not happy when we see comparing some distinct fields instead of ALL content (except some cases when it's really necessary). So, please, compare the whole result with the expected value. And it could be done in 1-2 lines of code, which is perfect for unit test. Fix same or similar issues in ALL places

src/test/java/core/basesyntax/service/FileWriterImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+
+class FileWriterImplTest {
+    private final FileWriter fileWriter = new FileWriterImpl();
+
+    @Test
+    void write_validFile_ok() throws IOException {
+        Path tempFile = Files.createTempFile("testOutput", ".csv");
+        String content = "fruit,quantity\nbanana,10";
```
sarakhmen: use system.lineseparator instead

src/test/java/core/basesyntax/service/FileWriterImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+    void write_validData_writesFile() throws IOException {
+        Path tempFile = Files.createTempFile("test", ".csv");
+        try {
+            FileWriter fileWriter = new FileWriterImpl();
+            fileWriter.write("banana,10", tempFile.toString());
+
+            String result = Files.readString(tempFile);
+            assertEquals("banana,10", result);
+        } finally {
+            Files.deleteIfExists(tempFile);
```
sarakhmen, line 43: do you really need to delete temp file? I'm not sure, it should be autodeleted even in case of exceptions if I'm not wrong

sarakhmen (changes requested):
let's improve your solution :)

Elena-Bruyako (approved):

PR #1224, opened 2025-11-06
Fix paths

AliakseiS97:


src/test/java/core/basesyntax/service/TransactionReaderServiceTest.java
```diff
@@ -0,0 +1,41 @@
...
+        ReaderService readerService = path -> List.of("b,apple,235");
+        TransactionReaderService service =
+                new TransactionReaderService(readerService);
+        List<String> result = service.readTransactions("anyPath.csv");
+        Assertions.assertEquals(1, result.size());
+        Assertions.assertEquals("b,apple,235", result.get(0));
+    }
+
+    @Test
+    public void readTransactions_validPath_ok() {
```
mateuszwojtkowiak, line 28: This test does not use any method from TransactionReaderService class so you can either delete it or modify like that: 
```@Test
public void readTransactions_multipleLines_ok() {
    ReaderService readerService = path -> List.of("b,apple,235", "s,banana,100");
    TransactionReaderService service = new TransactionReaderService(readerService);
    List<String> result = service.readTransactions("file.csv");
    Assertions.assertEquals(2, result.size());
    Assertions.assertTrue(result.contains("b,apple,235"));
    Assertions.assertTrue(result.contains("s,banana,100"));
}

mateuszwojtkowiak (approved):
Great job! I found that one of your tests is not using any method so please look at my comment.

PR #1264, opened 2026-01-11
wrote tests for fruit shop

sannchous:


src/test/java/resources/finalReportTest.csv
```diff
@@ -0,0 +1,3 @@
+fruit,quantity
```
Elena-Bruyako, line 1: resources folder should be placed in `src/test/resources`

Elena-Bruyako (approved):
Good job! One comment for the future

PR #1269, opened 2026-02-04
jv-fruit-shop-tests

jacekm91fidzinski-cmd:


mateuszwojtkowiak (changes requested):
Your code does not pass the tests. Please run `mvn clean package` locally to see what needs to be fixed. Push the project to GitHub again after all checks pass.

mateuszwojtkowiak (changes requested):
Your project doesn’t build because it contains only tests.
According to the task requirements, you must copy your solution from the previous task into src/main/java/core/basesyntax.
Right now your tests try to run code that doesn’t exist, so the build fails.

Before pushing your work to GitHub, always run mvn clean package locally.
Please read all requirements here: https://github.com/[mate-academy/jv-fruit-shop-tests](https://github.com/mate-academy/jv-fruit-shop-tests)

src/main/test/java/basesyntax/service/DataConverterImplTest.java
```diff
@@ -0,0 +1,29 @@
+package basesyntax.service;
```
slade13: Your test sources are placed here:

- `src/main/test/java/...`
- `src/main/test/resources/...`

They must be:

- `src/test/java/...`
- `src/test/resources/...`

Move the whole `src/main/test` directory to `src/test`.

src/main/test/java/basesyntax/service/DataConverterImplTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import basesyntax.model.FruitTransaction;
+import basesyntax.service.impl.DataConverterImpl;
```
slade13: Your production code packages are like:

- core.basesyntax.service.impl.DataConverterImpl
- core.basesyntax.db.Storage
- etc.

But your tests use:

- package basesyntax.service;
- imports like basesyntax.model.FruitTransaction, etc.

This will not compile.

src/main/test/java/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,53 @@
...
+class ShopServiceImplTest {
+
+    private ShopService shopService;
+
+    @BeforeEach
+    void setUp() {
+        // 🔥 KLUCZOWE – bez tego test zawsze będzie failował
+        Storage.clear();
+
+        Map<FruitTransaction.Operation, OperationHandler> handlers = new HashMap<>();
```
slade13: You build the full map of handlers + real OperationStrategyImpl + real operation handlers. That means you’re effectively testing strategy + handlers + service together. Task requirement explicitly says: don’t test strategy in service; you can use a map with only one handler or use a stub.

src/test/java/core/basesyntax/service/FileReaderImplTest.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class FileReaderImplTest {
+
+    @Test
+    void read_validFile_ok() {
+        FileReader reader = new FileReaderImpl();
+
+        List<String> result = reader.read("src/test/resources/input.csv");
```
slade13, line 20: Don't hardcode paths in your code - this breaks easily depending on working directory. Use classpath resource:
```
Path path = Paths.get(getClass().getClassLoader().getResource("input.csv").toURI());
List<String> lines = reader.read(path.toString());
```

slade13 (changes requested):
There are missing tests for:

- FileWriterImplTest
- OperationStrategyImplTest

Please review my comments and send your task again for a review.

slade13 (approved):
Good job!

PR #1270, opened 2026-02-12
hw-fruit-shop-tests is verified success

JasonTaranenko:


src/main/resources/valid_data.csv
```diff
@@ -0,0 +1,3 @@
+type,fruit,quantity
```
Elena-Bruyako, line 1: for the future, text files should be in `test/resources` folder

Elena-Bruyako (approved):
Good job 
for the future, text files should be in `test/resources` folder

PR #1286, opened 2026-03-19
feature/fruit-shop-tests

Krupkolllia:


src/test/java/core/basesyntax/service/impl/CsvFileWriterTest.java
```diff
@@ -0,0 +1,60 @@
...
+    void write_dataIsNull_notOk() {
+        assertThrows(NullPointerException.class, () ->
+                csvFileWriter.write(null, TEST_FILE_PATH));
+    }
+
+    @Test
+    void write_filePathIsNull_notOk() {
+        assertThrows(NullPointerException.class, () ->
+                csvFileWriter.write(TEST_FILE_CONTENT, null));
+    }
```
fetis-off:
```suggestion
```
I see you are testing cases for some NULL inputs and expecting that method will throw NPE. It's an obvious bahaviour you should not test it. Or create custom exception while NULL validation on your classes to test them will thow and it would be more logically and cleaner.

The same in other places where you have 
` assertThrows(NullPointerException.class, () ->`

src/test/java/core/basesyntax/util/impl/CsvShopReportGeneratorTest.java
```diff
@@ -0,0 +1,48 @@
...
+class CsvShopReportGeneratorTest {
+    private static final String TEST_REPORT_HEADER = "fruit,quantity";
+
+    private Storage shopStorage;
+    private final ReportGenerator reportGenerator = new CsvShopReportGenerator();
+
+    @BeforeEach
+    void setUp() {
+        shopStorage = new ShopStorage();
+    }
```
fetis-off: It would be better to create storage once  and clear it after each test.
The same in other places

fetis-off (changes requested):
Good work. Please, see comments

fetis-off (approved):
Good one!

PR #1288, opened 2026-03-24
add test for service implementations

chupa-ilona:


src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,52 @@
+package core.basesyntax;
```
Elena-Bruyako: package structure should be the same as in main code

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+    @Test
+    public void testConvertToTransaction_validData_OK() {
+        List<String> input = new ArrayList<>();
+        input.add("type,fruit,quantity");
+        input.add("b,apple,10");
+        input.add("s,orange,5");
+        input.add("p,banana,15");
+        input.add("r,orange,5");
+        List<FruitTransaction> transactions = converter.convertToTransaction(input);
+        Assertions.assertEquals(4, transactions.size());
```
Elena-Bruyako: use static imports
check all places

Elena-Bruyako (changes requested):
add test for all operations 
You have two resources folders, it should be one

Elena-Bruyako (changes requested):
you should have separate test class for each type of strategy with positive and negative test cases
Use static imports everywhere
Don't forget to clear Storage after each tests

fetis-off (approved):
Good work! Approved

fetis-off (changes requested):
Sorry, I did not noticed that build failed (there is a checkstyle errors)
Please fix build

PR #1291, opened 2026-03-30
Solved task with tests to previous task with SOLID

neuklearna:


src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,28 @@
...
+                "b,banana,20",
+                "b,apple,100",
+                "s,banana,100",
+                "p,banana,13",
+                "r,apple,10",
+                "p,apple,20",
+                "p,banana,5",
+                "s,banana,50"
+        );
+        List<String> read = fileReaderImpl.read("reportToRead.csv");
```
slade13: `FileReaderImplTest `uses a file outside `src/test/resources` - `read("reportToRead.csv")` reads a file from the project root, not from `src/test/resources`. This does not match the task requirement for test files location.

src/test/java/core/basesyntax/FileWriterImplTest.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+
+public class FileWriterImplTest {
+
+    @Test
+    void write_validContent_ok() throws IOException {
+        FileWriterImpl fileWriterImpl = new FileWriterImpl();
+        fileWriterImpl.write("fruit,quantity\napple,100\n", "testOutput.csv");
```
slade13: `FileWriterImplTest `writes output to the project root; it creates `testOutput.csv` in the root directory.
Test files should be kept under `src/test/resources`, and test output should not pollute the project root. For writing files inside your test use `@TempDir`.

src/test/java/core/basesyntax/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.ReportGenerator;
+import core.basesyntax.service.impl.ReportGeneratorImpl;
+import core.basesyntax.service.strategy.BalanceOperation;
+import org.junit.jupiter.api.Test;
+
+public class ReportGeneratorImplTest {
```
slade13, line 13: `ReportGeneratorImplTest `is not isolated, it uses `BalanceOperation `to prepare data for `ReportGeneratorImpl`.
According to the task, when testing report generation, storage should be filled directly instead of using another service/handler.

src/test/java/core/basesyntax/ShopServiceImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+import core.basesyntax.service.impl.OperationStrategyImpl;
+import core.basesyntax.service.impl.ShopServiceImpl;
+import core.basesyntax.service.strategy.BalanceOperation;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class ShopServiceImplTest {
```
slade13, line 18: `ShopServiceImplTest `is not isolated, it uses the real `BalanceOperation `in service tests.
This mixes `ShopServiceImpl `testing with handler logic. The task explicitly says strategy/handler/service tests should stay separate.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+    void convertToTransaction_nullData_notOk() {
+        List<String> data = null;
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(data));
+    }
+
+    @Test
+    void setters_validData_ok() {
```
slade13: `ShopServiceImplTest `is not isolated, it uses the real `BalanceOperation` in service tests.
This mixes `ShopServiceImpl` testing with handler logic. The task explicitly says strategy/handler/service tests should stay separate.

src/test/java/core/basesyntax/DataConventerImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.impl.DataConverterImpl;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class DataConventerImplTest {
```
slade13: DataConventerImplTest class name is misspelled:

```suggestion
public class DataConverterImplTest {
```

src/test/java/core/basesyntax/ReturnOperationTest.java
```diff
@@ -0,0 +1,48 @@
...
+                FruitTransaction.Operation.RETURN);
+
+        balanceOperation.handle(transaction);
+        returnOperation.handle(transaction2);
+
+        assertEquals(105, storage.getStorage().get("apple"));
+    }
+
+    @Test
+    void handle_emptyStorage_NotOk() {
```
slade13:
```suggestion
    void handle_emptyStorage_notOk() {
```

src/test/java/core/basesyntax/ShopServiceImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    void process_emptyList_ok() {
```
slade13, line 21: This test has no assertions, it can pass even with some errors during tests or empty methods.

src/test/java/core/basesyntax/DataConventerImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+        List<String> data = List.of("quantity,fruit,operation",
+                "r,apple,100",
+                "p,apple,20",
+                "b,apple,30",
+                "s,apple,10");
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        List<FruitTransaction> fruitTransactions = dataConverter.convertToTransaction(data);
+
+        assertEquals(4, fruitTransactions.size());
```
slade13: This test checks only list size, not whether operations were mapped correctly.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        List<FruitTransaction> result = dataConverter.convertToTransaction(data);
+
+        assertEquals(transaction, result.get(0));
+    }
+
+    @Test
+    void convertToTransaction_skipHeader_ok() {
```
slade13: This test mostly duplicates the first test and does not verify enough.

slade13 (changes requested):
Good job on creating tests for your Fruit Shop! 

There are still few things which require changes, additionally several tests use `assertThrows(RuntimeException.class, ...)` for null input.
That passes even for accidental `NullPointerException`, so the test may hide bad implementation details instead of checking intended behavior.

src/test/java/core/basesyntax/ShopServiceImplTest.java
neuklearna: jak moge to rozwiazac? moge posluzyc sie tutaj mockito czy jest jakies inne rozwiazanie?

src/test/java/core/basesyntax/FileReaderImplTest.java
slade13: `FileReaderImplTest `still does not use `src/test/resources` properly, this depends on the working directory instead of loading the resource from test resources in a reliable way.
This
```
fileReader.read("reportToRead.csv");
```
works only when the working directory happens to be set in a way that the file is found.
A test should read the file from `src/test/resources` reliably. Better approach is:
```
URL resource = getClass().getClassLoader().getResource("reportToRead.csv");
String path = Paths.get(Objects.requireNonNull(resource).toURI()).toString();

List<String> actual = fileReader.read(path);
```

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+    void convertToTransaction_nullData_notOk() {
+        List<String> data = null;
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(data));
+    }
+
+    @Test
+    void setters_validData_ok() {
```
slade13: `setters_validData_ok()` tests `FruitTransaction `setters, not `DataConverterImpl`. This should not be inside converter tests.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+        List<String> data = List.of("quantity,fruit,operation",
+                "b,apple,100",
+                "p,apple,20",
+                "b,apple,30",
+                "s,apple,10");
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        List<FruitTransaction> fruitTransactions = dataConverter.convertToTransaction(data);
+
+        assertEquals(FruitTransaction.Operation.BALANCE, fruitTransactions.get(0).getOperation());
```
slade13, line 55: This verifies only the first operation. It should verify all parsed entries, not just that index `0` is `BALANCE`.

src/test/java/core/basesyntax/ReturnOperationTest.java
slade13: Please update the name to be:

```suggestion
    void handle_emptyStorage_notOk() {
```

src/test/java/core/basesyntax/ReturnOperationTest.java
```diff
@@ -0,0 +1,48 @@
...
+        Storage storage = new Storage();
+        ReturnOperation returnOperation = new ReturnOperation(storage);
+        FruitTransaction transaction2 = new FruitTransaction(5, "apple",
+                FruitTransaction.Operation.RETURN);
+
+        assertThrows(RuntimeException.class, () -> returnOperation.handle(transaction2));
+    }
+
+    @Test
+    void handle_nullTransaction_NotOk() {
```
slade13:
```suggestion
    void handle_nullTransaction_notOk() {
```

src/test/java/core/basesyntax/ShopServiceImplTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+        FruitTransaction transaction = new FruitTransaction(100, "apple",
+                FruitTransaction.Operation.BALANCE);
+        shopService.process(List.of(transaction));
+
+        assertTrue(fakeOperationHandler.isWasCalled());
+    }
+
+    @Test
+    void process_nullList_NotOk() {
```
slade13:
```suggestion
    void process_nullList_notOk() {
```

src/test/java/core/basesyntax/BalanceOperationTest.java
```diff
@@ -0,0 +1,66 @@
...
+        // GIVEN — przygotuj dane
+        Storage storage = new Storage();
+        BalanceOperation balanceOperation = new BalanceOperation(storage);
+        FruitTransaction transaction = new FruitTransaction(100, "banana",
+                FruitTransaction.Operation.BALANCE);
+
+        // WHEN — wywołaj metodę
+        balanceOperation.handle(transaction);
+
+        // THEN — sprawdź wynik
```
slade13: Comments are left in Polish, not a functional error, but for review quality and consistency in this project, English comments are preferred.

```suggestion
        // GIVEN
        Storage storage = new Storage();
        BalanceOperation balanceOperation = new BalanceOperation(storage);
        FruitTransaction transaction = new FruitTransaction(100, "banana",
                FruitTransaction.Operation.BALANCE);

        // WHEN
        balanceOperation.handle(transaction);

        // THEN
```

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.service.impl.FileReaderImpl;
+import java.io.IOException;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class FileReaderImplTest {
```
slade13, line 17: `FileReaderImplTest `has only one positive scenario. For this task, tests should try to cover more than only the happy path, especially if the service can fail on missing file / invalid path.

slade13 (changes requested):
This version is much better than the previous one, but there are still some changes required before approval. The biggest remaining problems are:

- improper test resource handling in `FileReaderImplTest`
- unrelated setter test inside `DataConverterImplTest`
- weak assertions in converter tests - some tests use assertThrows(RuntimeException.class, ...). That can hide accidental failures and makes tests less precise.
- inconsistent `notOk `naming

Moreover all these comments were stated in previous code review, before sending your code for a review again please make sure everything is fixed and updated.

src/test/java/core/basesyntax/BalanceOperationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.strategy.BalanceOperation;
+import org.junit.jupiter.api.Test;
+
+public class BalanceOperationTest {
+
+    @Test
+    void handle_validTransaction_ok() {
+        Storage storage = new Storage();
```
mateuszwojtkowiak: Both `Storage storage = new Storage();` and `BalanceOperation balanceOperation = new BalanceOperation(storage);` are repeated in every test.  
Since these objects are required for each test and are not static, this is correct, but you could extract them into a `@BeforeEach` method to reduce duplication and improve readability.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        List<FruitTransaction> result = dataConverter.convertToTransaction(data);
+
+        assertEquals(transaction, result.get(0));
+    }
+
+    @Test
+    void convertToTransaction_skipHeader_ok() {
```
mateuszwojtkowiak: This test duplicates the logic of `convertToTransaction_validData_ok` and does not verify that the header is actually skipped.  
A proper header-skip test should assert that the header line is ignored rather than repeating the same assertions.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    void convertToTransaction_validData_ok() {
+        List<String> data = List.of("quantity,fruit,operation", "b,apple,100");
+        FruitTransaction transaction = new FruitTransaction(100, "apple",
+                FruitTransaction.Operation.BALANCE);
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        List<FruitTransaction> result = dataConverter.convertToTransaction(data);
```
mateuszwojtkowiak, line 27: The test assumes that the result list contains at least one element but does not assert its size.  
If the converter returns an empty list, the test will fail with IndexOutOfBoundsException instead of a meaningful assertion failure.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+        FruitTransaction takeFirst = result.get(0);
+        assertEquals(transaction, takeFirst);
+    }
+
+    @Test
+    void convertToTransaction_allOperations_ok() {
```
mateuszwojtkowiak: The test name suggests that all operations are tested, but only BALANCE is used.  
This is misleading and should be corrected either by renaming the test or by actually testing all operations.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.impl.DataConverterImpl;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class DataConverterImplTest {
+
+    @Test
+    void convertToTransaction_validData_ok() {
+        List<String> data = List.of("quantity,fruit,operation", "b,apple,100");
```
mateuszwojtkowiak: The test uses the header "quantity,fruit,operation", but the expected CSV format is "type,fruit,quantity".  
This mismatch may hide parsing issues because the test does not reflect the actual input structure.  
Please update the test data to match the correct CSV format.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+        FruitTransaction transaction = new FruitTransaction(100, "apple",
+                FruitTransaction.Operation.BALANCE);
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
```
mateuszwojtkowiak: `DataConverterImpl` is instantiated in every test.  
Since it has no internal state and is required for each test, you can extract this into a `@BeforeEach` method to reduce duplication and improve readability.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+        List<String> data = List.of("quantity,fruit,operation", "x,apple,100");
+
+        DataConverterImpl dataConverter = new DataConverterImpl();
+
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(data));
+    }
+
+    @Test
+    void convertToTransaction_nullData_notOk() {
+        List<String> data = null;
```
mateuszwojtkowiak: There is no need to create a `List<String>` variable here since the test passes `null` directly.  
You can simplify the test by calling `convertToTransaction(null)` in the assertion without creating an unused variable.

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,39 @@
...
+                "p,apple,20",
+                "p,banana,5",
+                "s,banana,50"
+        );
+
+        URL resource = getClass().getClassLoader().getResource("reportToRead.csv");
+        String path = Paths.get(Objects.requireNonNull(resource).toURI()).toString();
+
+        List<String> read = fileReaderImpl.read(path);
+        assertEquals(strings, read);
```
mateuszwojtkowiak: The test currently compares the entire file content line‑by‑line, which makes it fragile.  
Any small formatting change in the CSV file (extra newline, different line endings, reordered lines) would cause the test to fail even though the reader works correctly.  
A more robust approach is to assert only that the file was read successfully and that key lines (e.g., the header) are present, rather than comparing the whole file verbatim.
It could look like that:
```
@Test
void read_validFile_ok() throws Exception {
    FileReaderImpl fileReaderImpl = new FileReaderImpl();

    URL resource = getClass().getClassLoader().getResource("reportToRead.csv");
    String path = Paths.get(Objects.requireNonNull(resource).toURI()).toString();

    List<String> read = fileReaderImpl.read(path);

    assertFalse(read.isEmpty());
    assertEquals("type,fruit,quantity", read.get(0));
}

File: src/test/java/core/basesyntax/FileWriterImplTest.java [hunk 25]
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class FileWriterImplTest {
+
+    @TempDir
+    private Path tempDir;
+
+    @Test
+    void write_validContent_ok() throws IOException {
```
mateuszwojtkowiak, line 18: The test compares the entire file content as a single string, including the exact newline characters.  
This makes the test fragile, especially across different operating systems where line separators may differ (`\n` vs `\r\n`).  
A more robust approach is to read the file line‑by‑line and assert individual lines instead of comparing the whole raw string.
It could look like that:
```
@Test
void write_validContent_ok() throws IOException {
    FileWriterImpl fileWriterImpl = new FileWriterImpl();
    String content = "fruit,quantity\napple,100\n";

    Path output = tempDir.resolve("testOutput.csv");
    fileWriterImpl.write(content, output.toString());

    List<String> lines = Files.readAllLines(output);

    assertEquals("fruit,quantity", lines.get(0));
    assertEquals("apple,100", lines.get(1));
}

File: src/test/java/core/basesyntax/FruitTransactionImplTest.java [hunk 26]
```diff
@@ -0,0 +1,20 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.model.FruitTransaction;
+import org.junit.jupiter.api.Test;
+
+public class FruitTransactionImplTest {
+
+    @Test
+    void setters_validData_ok() {
```
mateuszwojtkowiak, line 11: This test modifies three fields but asserts only one of them, so it does not actually verify the behavior of all setters being used.  
Additionally, testing simple getters/setters in a data model is usually unnecessary unless they contain custom logic.  
You can expand this to assert all modified fields (quantity, fruit and operation).

src/test/java/core/basesyntax/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+import core.basesyntax.service.impl.OperationStrategyImpl;
+import core.basesyntax.service.strategy.BalanceOperation;
+import core.basesyntax.service.strategy.PurchaseOperation;
+import core.basesyntax.service.strategy.ReturnOperation;
+import core.basesyntax.service.strategy.SupplyOperation;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class OperationStrategyImplTest {
```
mateuszwojtkowiak, line 20: Each test duplicates the same setup logic (creating Storage, the handler map, and the strategy).  
This makes the tests harder to maintain and obscures the actual behavior being tested.  
You can extract the common setup into a `@BeforeEach` method.

src/test/java/core/basesyntax/PurchaseOperationTest.java
```diff
@@ -0,0 +1,60 @@
...
+        FruitTransaction transactionPurchase = new FruitTransaction(25, "apple",
+                FruitTransaction.Operation.PURCHASE);
+
+        balanceOperation.handle(transactionBalance);
+
+        assertThrows(RuntimeException.class, () -> operation.handle(transactionPurchase));
+    }
+
+    @Test
+    void handle_zeroQuantity_notOk() {
```
mateuszwojtkowiak, line 52: The test assumes that a purchase with quantity 0 should be accepted and leave the storage unchanged.  
However, a transaction with zero quantity is invalid input and should normally result in an exception rather than being processed.  
This test may hide incorrect behavior in the PurchaseOperation implementation.

src/test/java/core/basesyntax/PurchaseOperationTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.strategy.BalanceOperation;
+import core.basesyntax.service.strategy.PurchaseOperation;
+import org.junit.jupiter.api.Test;
+
+public class PurchaseOperationTest {
```
mateuszwojtkowiak, line 13: Each test repeats the same setup (Storage, BalanceOperation, PurchaseOperation).  
This duplication makes the tests harder to maintain.  
You can extract the common setup into a `@BeforeEach` method.

src/test/java/core/basesyntax/ReturnOperationTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.strategy.BalanceOperation;
+import core.basesyntax.service.strategy.ReturnOperation;
+import org.junit.jupiter.api.Test;
+
+public class ReturnOperationTest {
```
mateuszwojtkowiak, line 13: The setup (Storage, BalanceOperation, ReturnOperation) is duplicated across tests and can be extracted into a `@BeforeEach` method for better maintainability.

src/test/java/core/basesyntax/SupplyOperationTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.strategy.BalanceOperation;
+import core.basesyntax.service.strategy.SupplyOperation;
+import org.junit.jupiter.api.Test;
+
+public class SupplyOperationTest {
```
mateuszwojtkowiak, line 13: The setup (Storage, SupplyOperation, BalanceOperation) is duplicated across tests.  
Extracting this into a `@BeforeEach` method would improve readability and maintainability.

src/test/java/core/basesyntax/SupplyOperationTest.java
```diff
@@ -0,0 +1,67 @@
...
+                FruitTransaction.Operation.SUPPLY);
+
+        operation2.handle(startValue);
+        operation.handle(transaction);
+
+        assertEquals(120, storage.getStorage().get("banana"));
+    }
+
+    @Test
+    void handle_emptyStorage_ok() {
```
mateuszwojtkowiak: The test name `handle_emptyStorage_ok` suggests a successful scenario,  
but the test actually expects an exception.  
Consider renaming it to reflect the expected failure.

mateuszwojtkowiak (changes requested):

src/test/java/core/basesyntax/ShopServiceImplTest.java
slade13: Tak, możesz użyć mockito. Takie pytania najlepiej zadawaj na naszym czacie w Mate.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+
+    public static void main(String[] arg) throws IOException {
+
+        Storage storage = new Storage();
+        // 1. Czytanie danych z pliku CSV
```
slade13: Remove these comments or translate them to english.

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,30 @@
...
+        URL resource = getClass().getClassLoader().getResource("reportToRead.csv");
+        String path = Paths.get(Objects.requireNonNull(resource).toURI()).toString();
+
+        List<String> read = fileReaderImpl.read(path);
+
+        assertFalse(read.isEmpty());
+        assertEquals("operation,fruit,quantity", read.get(0));
+    }
+
+    //My task passed with 80% line and that's why i didn't add test with negaitve result
```
slade13: Code coverage is not everything what is expected during unit test creation. We want to have all the business cases covered and additionally we want to have more than 80% of code coverage in tests.

Negative scenarios have to be tested, for example we want to testify if someone is an adult, we write the test checking if our method `isAdult(...)` will return `true` for age `20`, but what in case if our code will also return `true` for values like `10, 16, 17`?

Please don't make comments like that in your code, instead open discussion on Mate Academy chat.

src/test/java/core/basesyntax/SupplyOperationTest.java
slade13: Like previous mentor said, this test expects exception but the name states differently: `handle_emptyStorage_ok` - please fix such things before requesting review of your code.

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @BeforeEach
+    void setup() {
+        dataConverter = new DataConverterImpl();
+    }
+
+    @Test
+    void convertToTransaction_validData_ok() {
+        List<String> data = List.of("operation,fruit,quantity", "b,apple,100");
+        FruitTransaction transaction = new FruitTransaction(100, "apple",
```
slade13, line 24: You've created an expected transaction object, but it's not compared with the actual result fields.
Right now it only checks size of the result.

slade13 (changes requested):
Please review my comments and update your code accordingly, you are really close to really good solution.

src/test/java/core/basesyntax/BalanceOperationTest.java
```diff
@@ -0,0 +1,64 @@
...
+        FruitTransaction transaction = new FruitTransaction(100, "banana",
+                FruitTransaction.Operation.BALANCE);
+
+        balanceOperation.handle(transaction);
+
+        assertEquals(100, storage.getStorage().get("banana"));
+    }
+
+    @Test
+    void handle_twoFruits_ok() {
```
mateuszwojtkowiak: Test name `handle_twoFruits_ok` is not fully descriptive.
Should follow the required convention more precisely, e.g.:
`handle_twoDifferentFruits_ok`

src/test/java/core/basesyntax/BalanceOperationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Storage;
+import core.basesyntax.service.strategy.BalanceOperation;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class BalanceOperationTest {
```
mateuszwojtkowiak, line 12: Try to cover more negative‑path tests like:
- empty fruit name
- negative quantity

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+        List<String> data = List.of("operation,fruit,quantity",
+                "b,apple,100",
+                "p,apple,25",
+                "s,apple,30",
+                "r,apple,15");
+
+        List<FruitTransaction> fruitTransactions = dataConverter.convertToTransaction(data);
+
+        assertEquals(4, fruitTransactions.size());
+        assertEquals(FruitTransaction.Operation.BALANCE, fruitTransactions.get(0).getOperation());
```
mateuszwojtkowiak, line 55: The test currently verifies only the operation field for each parsed transaction.
Since DataConverterImpl parses three fields (operation, fruit, quantity),
the test should also assert the fruit name and quantity for every entry.

Otherwise, bugs in parsing fruit or quantity would not be detected.

Example of a more complete assertion:

```
assertEquals(FruitTransaction.Operation.BALANCE, fruitTransactions.get(0).getOperation());
assertEquals("apple", fruitTransactions.get(0).getFruit());
assertEquals(100, fruitTransactions.get(0).getQuantity());

```

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+        assertEquals(FruitTransaction.Operation.PURCHASE, fruitTransactions.get(1).getOperation());
+        assertEquals(FruitTransaction.Operation.SUPPLY, fruitTransactions.get(2).getOperation());
+        assertEquals(FruitTransaction.Operation.RETURN, fruitTransactions.get(3).getOperation());
+    }
+
+    @Test
+    void convertToTransaction_invalidQuantity_notOk() {
+        List<String> data = List.of("operation,fruit,quantity", "b,apple,abc");
+        DataConverterImpl dataConverter = new DataConverterImpl();
```
mateuszwojtkowiak: Creating dataConverter here is reduntant, because it is created in `setup()` method.

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Objects;
+import org.junit.jupiter.api.Test;
+
+public class FileReaderImplTest {
```
mateuszwojtkowiak, line 17: Good point — the test suite covers the happy path, empty file and missing file,
but it doesn't validate behavior when the file exists but contains invalid or unreadable content.

It would be useful to add a negative test for a file with incorrect format or corrupted data,
e.g. a file containing only whitespace or binary garbage, to ensure FileReaderImpl
fails gracefully on invalid input.

Example:
```
@Test
void read_fileWithInvalidContent_notOk() throws Exception {
    // Your business logic here.
}

```

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+    @Test
+    void read_emptyFile_notOK() throws Exception {
+        Path emptyFile = Files.createTempFile("empty",".csv");
+        emptyFile.toFile().deleteOnExit();
+        assertThrows(RuntimeException.class, () -> fileReaderImpl.read(emptyFile.toString()));
+    }
+
+    @Test
+    void read_fileNotExist_notOk() {
+        assertThrows(IOException.class, () -> new FileReaderImpl().read("notExist.csv"));
```
mateuszwojtkowiak: Since you create `fileReaderImpl` field under class, you don't need to create a new one. Just use the existing object.

src/test/java/core/basesyntax/FileWriterImplTest.java
mateuszwojtkowiak, line 18: The current test compares the entire file content as a single raw string, including
the exact newline characters. This makes the test fragile across different operating
systems, because line separators may differ (\n vs \r\n).

A more robust approach is to read the file line‑by‑line and assert individual lines.
This avoids OS‑dependent failures and makes the test more stable.

Example:

```
@Test
void write_validContent_ok() throws IOException {
    FileWriterImpl fileWriterImpl = new FileWriterImpl();
    String content = "fruit,quantity\napple,100\n";

    Path output = tempDir.resolve("testOutput.csv");
    fileWriterImpl.write(content, output.toString());

    List<String> lines = Files.readAllLines(output);

    assertEquals("fruit,quantity", lines.get(0));
    assertEquals("apple,100", lines.get(1));
}
```

src/test/java/core/basesyntax/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class OperationStrategyImplTest {
+
+    private Storage storage;
+
+    @BeforeEach
+    void setUp() {
```
mateuszwojtkowiak, line 27: Each test repeats the same setup logic (creating the handler map and the strategy).
Extracting this into a `@BeforeEach` method would reduce duplication and make the tests
easier to read and maintain. The individual tests would then focus only on the behavior
being verified instead of recreating the same boilerplate.
Example:
```
private Map<FruitTransaction.Operation, OperationHandler> handler;
private OperationStrategy strategy;

@BeforeEach
void setUp() {
    storage = new Storage();
    handler = new HashMap<>();
    strategy = new OperationStrategyImpl(handler);
}

```

src/test/java/core/basesyntax/PurchaseOperationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class PurchaseOperationTest {
+
+    private Storage storage;
+
+    @BeforeEach
+    void setUp() {
+        storage = new Storage();
```
mateuszwojtkowiak, line 21: In order to avoid creating duplicated code, add necessary objects to `setUp()` method and use them.

```
private Storage storage;
    private BalanceOperation balanceOperation;
    private PurchaseOperation purchaseOperation;

    @BeforeEach
    void setUp() {
        storage = new Storage();
        balanceOperation = new BalanceOperation(storage);
        purchaseOperation = new PurchaseOperation(storage);
    }
```

src/test/java/core/basesyntax/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ReportGeneratorImplTest {
+
+    private Storage storage;
+
+    @BeforeEach
+    void setUp() {
```
mateuszwojtkowiak, line 18: You can enhance setUp() to avoid duplication.
```
private Storage storage;
    private ReportGenerator reportGenerator;

    @BeforeEach
    void setUp() {
        storage = new Storage();
        reportGenerator = new ReportGeneratorImpl(storage);
    }
```

src/test/java/core/basesyntax/ReturnOperationTest.java
```diff
@@ -0,0 +1,53 @@
...
+import core.basesyntax.service.strategy.ReturnOperation;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ReturnOperationTest {
+
+    private Storage storage;
+
+    @BeforeEach
+    void setUp() {
```
mateuszwojtkowiak, line 20: You can enhance setUp() method to avoid duplication.
```
private Storage storage;
    private ReturnOperation returnOperation;
    private BalanceOperation balanceOperation;

    @BeforeEach
    void setUp() {
        storage = new Storage();
        returnOperation = new ReturnOperation(storage);
        balanceOperation = new BalanceOperation(storage);
    
```

src/test/java/core/basesyntax/SupplyOperationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import core.basesyntax.service.strategy.SupplyOperation;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class SupplyOperationTest {
+
+    private Storage storage;
+
+    @BeforeEach
+    void setUp() {
```
mateuszwojtkowiak, line 20: Avoid duplication by enhancing your `setUp()` method.
```
private Storage storage;
    private SupplyOperation supplyOperation;
    private BalanceOperation balanceOperation;

    @BeforeEach
    void setUp() {
        storage = new Storage();
        supplyOperation = new SupplyOperation(storage);
        balanceOperation = new BalanceOperation(storage);
    }
```

mateuszwojtkowiak (changes requested):
Good job! You are getting closer to finishing this task. Please read my comments and apply necessary changes.

src/test/java/core/basesyntax/SupplyOperationTest.java
```diff
@@ -0,0 +1,68 @@
...
+    void handle_emptyStorage_notOk() {
+        FruitTransaction transaction = new FruitTransaction(100, "apple",
+                FruitTransaction.Operation.SUPPLY);
+
+        assertThrows(RuntimeException.class, () -> operation.handle(transaction));
+    }
+
+    @Test
+    void handle_multipleSupplies_ok() {
+        FruitTransaction balanseOperationValue = new FruitTransaction(20, "apple",
```
slade13, line 49: Small typo:
```suggestion
        FruitTransaction balanceOperationValue = new FruitTransaction(20, "apple",
```

src/test/java/core/basesyntax/OperationStrategyImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+        SupplyOperation supplyOperation = new SupplyOperation(storage);
+        handler.put(FruitTransaction.Operation.SUPPLY, supplyOperation);
+
+        assertInstanceOf(SupplyOperation.class,
+                strategy.getHandler(FruitTransaction.Operation.SUPPLY));
+    }
+
+    @Test
+    void getHandler_return_ok() {
+        ReturnOperation returnOperation = new ReturnOperation(storage);;
```
slade13, line 62: Double semicolon:

```suggestion
        ReturnOperation returnOperation = new ReturnOperation(storage);
```

src/test/java/core/basesyntax/FileReaderImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+        String path = Paths.get(Objects.requireNonNull(resource).toURI()).toString();
+
+        List<String> read = fileReaderImpl.read(path);
+
+        assertFalse(read.isEmpty());
+        assertEquals("operation,fruit,quantity", read.get(0));
+    }
+
+    @Test
+    void read_emptyFile_notOK() throws Exception {
```
slade13, line 33: Make your names consistent:

```suggestion
    void read_emptyFile_notOk() throws Exception {
```

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    void convertToTransaction_unknownOperation_notOk() {
+        List<String> data = List.of("operation,fruit,quantity", "x,apple,100");
+
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(data));
+    }
+
+    @Test
+    void convertToTransaction_nullData_notOk() {
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(null));
```
slade13, line 76: Use more specific exception:

```suggestion
        assertThrows(NullPointerException.class, () -> dataConverter.convertToTransaction(null));
```

src/test/java/core/basesyntax/DataConverterImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertEquals(FruitTransaction.Operation.BALANCE, fruitTransactions.get(0).getOperation());
+        assertEquals(FruitTransaction.Operation.PURCHASE, fruitTransactions.get(1).getOperation());
+        assertEquals(FruitTransaction.Operation.SUPPLY, fruitTransactions.get(2).getOperation());
+        assertEquals(FruitTransaction.Operation.RETURN, fruitTransactions.get(3).getOperation());
+    }
+
+    @Test
+    void convertToTransaction_invalidQuantity_notOk() {
+        List<String> data = List.of("operation,fruit,quantity", "b,apple,abc");
+        assertThrows(RuntimeException.class, () -> dataConverter.convertToTransaction(data));
```
slade13, line 64:
```suggestion
        assertThrows(NumberFormatException.class, () -> dataConverter.convertToTransaction(data));
```

slade13 (approved):
There are still some parts in your code which would require better implementation/handling. In many places you throw `RuntimeException` instead of your custom unchecked exceptions which hides the actual behavior in your tests.

Besides that if you expect getting `NullPointerException` from your method then you should use `NullPointerException` in your test instead of `RuntimeException` which is too broad and hides what really happens. Sample places where it happens:

- ReturnOperation.handle() on empty storage -> NullPointerException
- ReturnOperation.handle(null) -> NullPointerException
- SupplyOperation.handle() on empty storage -> NullPointerException
- SupplyOperation.handle(null) -> NullPointerException
- ShopServiceImpl.process(null) -> NullPointerException

`PurchaseOperationTest.handle_zeroQuantity_notOk()` is also problematic for a bigger reason:

`PurchaseOperation.handle()` does not validate zero quantity
with current implementation, for "apple" not existing in storage, it fails because of:
`balance.get(fruit) < quantity`

which causes `NullPointerException` due to unboxing `null`

so the test passes only because it expects broad `RuntimeException`, not because zero quantity is truly handled.

Besides that I approve your PR but please review all the comments stated by mentors in this project and please make sure that you understand them and why they were written. In case of any doubts please contact us on Mate chat.

PR #1298, opened 2026-04-13
Develop

slawomir-pacek:


src/test/basesyntax/HelloWorldTest.java
mateuszwojtkowiak, line 1: HelloWorldTest class should be removed.

mateuszwojtkowiak (changes requested):
Right now `mvn clean package` does not run any tests at all, because the test classes are located outside the module’s source root. The root cause is an incorrect folder structure. The project must follow the standard Maven layout:

project/
 └── src/
      ├── main/
      │    └── java/
      │         └── basesyntax/
      │              └── ...
      └── test/
           └── java/
                └── basesyntax/
                     └── ...

The current `core` folder should be removed, and the code should be placed exactly as shown above. Without this structure, IntelliJ and Maven will not detect or compile the tests.

Additionally, the project is missing required dependencies and plugins from the reference `pom.xml`, including JaCoCo. A correct example can be found here:
https://github.com/mate-academy/jv-fruit-shop-tests/blob/main/pom.xml

After fixing the directory layout, the tests still do not pass Checkstyle and do not meet the required 80% line coverage.

PR #1305, opened 2026-04-20
copied files from previous task and tests added

ReduxR:


src/main/java/core/basesyntax/checker/Checker.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.checker;
+
+import core.basesyntax.db.Storage;
+
+public class Checker {
```
fetis-off: Very generic class name. Please call it more obvious to avoid misunderstanding in the code

src/main/java/core/basesyntax/checker/Checker.java
```diff
@@ -0,0 +1,33 @@
...
+    public void validateFruit(String fruit) {
+        if (fruit == null) {
+            throw new IllegalArgumentException("Fruit must not be null.");
+        }
+        if (!storage.contains(fruit)) {
+            throw new RuntimeException("Cannot find fruit: " + fruit);
+        }
+    }
+    
+    public void validateNewEntry(String fruit, int quantity) {
```
fetis-off:
```suggestion
    public void validateNewFruitEntry(String fruit, int quantity) {
```

src/test/java/core/basesyntax/dao/TransactionDaoTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class TransactionDaoTest {
+
+    private Storage storage;
+    private Checker checker;
+    private TransactionDao dao;
+
+    @BeforeEach
```
fetis-off, line 19:
```suggestion
    @BeforeAll
```
I am wondering why do we need create new instances of classes before each test?
The same in other places

src/test/java/core/basesyntax/handler/FakeTransactionDao.java
```diff
@@ -0,0 +1,48 @@
+package core.basesyntax.handler;
+
+import core.basesyntax.service.TransactionDao;
+import java.util.Map;
+
+public class FakeTransactionDao implements TransactionDao {
```
fetis-off: Could you explain what is the purpose of this class?

src/test/java/core/basesyntax/handler/ReturnHandlerTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.strategy.handler.ReturnHandler;
+import org.junit.jupiter.api.Test;
+
+public class ReturnHandlerTest {
+    @Test
+    public void handle_ReturnHandler_ok() {
```
fetis-off: Good test names follow:
method + expected behavior + condition
example: `handle_shouldAddFruitToDao_whenReturnOperation()`

fetis-off (changes requested):
Left comments

src/test/java/core/basesyntax/handler/FakeTransactionDao.java
ReduxR: Since I was trying to follow unit testing principles, I created a fake implementation of TransactionDao to isolate the handler logic and verify interactions. However, now I think that direct interaction with Storage would be more appropriate for this task

src/test/java/core/basesyntax/dao/TransactionDaoTest.java
ReduxR: I removed the static state from Storage, so each test now works with its own isolated instance
Because of that, BeforeEach remains the appropriate choice to ensure that tests gets an independent and fresh setup. Using BeforeAll would not provide good isolation level

src/test/java/core/basesyntax/handler/FakeTransactionDao.java
fetis-off: Yes, please remove it

src/main/java/core/basesyntax/checker/Checker.java
fetis-off: resolve it

src/test/java/core/basesyntax/dao/TransactionDaoTest.java
fetis-off: Good point. But if you have stateless components you can define it once, for example your Checker class  is independent from any state so it's just redundant creation of this object before each test.

fetis-off (changes requested):
Thank you for your replies! See comments

src/test/java/core/basesyntax/dao/TransactionDaoTest.java
ReduxR: my Checker is tied to the Storage instance. Since Storage is recreated per test for isolation, I think it is better if Checker(EntryValidator is a new name for it now) would be recreated as well to avoid referencing outdated state

fetis-off (approved):
Good

PR #1308, opened 2026-04-22
Jv fruit shop tests

Dimon1888:


src/reportToRead.csv
fetis-off, line 1: could you explain why do you have two exactly the same files?

fetis-off (approved):
Good one!

PR #1319, opened 2026-05-29
firstSolution

mate-student26:


src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    void validInputData_Ok() {
+        List<String> inputReport = List.of(
+                "fruit,quantity",
+                "b,banana,100",
+                "s,apple,50"
+        );
+
+        List<FruitTransaction> transactions = dataConverter.convertToTransaction(inputReport);
+        assertEquals(transactions.size(), 2);
```
mateuszwojtkowiak: The expected and actual values in `assertEquals` are reversed. 
The correct order is `assertEquals(expected, actual)`, so it should be:

`assertEquals(2, transactions.size());`

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+                "x,banana,100"
+        );
+
+        assertThrows(RuntimeException.class, () ->
+                dataConverter.convertToTransaction(inputReport)
+        );
+    }
+
+    @Test
+    void emptyList_NotOk() {
```
mateuszwojtkowiak: The test `emptyList_NotOk` is incorrect. According to the requirements, 
an empty input should result in an exception, not an empty list. 
This test should expect a RuntimeException instead of checking for an empty result.

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class DataConverterImplTest {
+    private final DataConverterImpl dataConverter = new DataConverterImpl();
+
+    @Test
+    void validInputData_Ok() {
```
mateuszwojtkowiak: The test name `validInputData_Ok` does not follow the required naming convention.  
It should start with the method under test, e.g.:

convertToTransaction_validInput_ok

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class DataConverterImplTest {
```
mateuszwojtkowiak, line 10: Consider adding a test case for `null` input.

mateuszwojtkowiak, line 10: The class uses `RuntimeException` for invalid input cases, which is too generic.  
Since the errors here are caused by incorrect or malformed input data,  
it is better to throw a more specific exception type such as `IllegalArgumentException`.

Using a precise exception improves readability, makes debugging easier,  
and clearly communicates that the issue is related to invalid method arguments.

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+
+public class FileReaderImplTest {
+    private final FileReaderImpl fileReaderImpl = new FileReaderImpl();
+
+    @Test
+    void read_validPath_Ok() {
```
mateuszwojtkowiak: This test does not verify the behavior of the `read` method.  
It only checks whether the file exists on disk, which is unrelated to the method logic.  
The test should assert the returned list content instead of checking file existence.

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,25 @@
...
+        String path = "src/test/java/resources_test/finalReport.csv";
+        fileReaderImpl.read(path);
+        assertTrue(Files.exists(Paths.get(path)));
+    }
+
+    @Test
+    void read_invalidPath_throwsException() {
+        assertThrows(RuntimeException.class, () ->
```
mateuszwojtkowiak: The test expects a generic `RuntimeException`, but the implementation  
should throw a more specific exception type (e.g. `IllegalArgumentException`).  
The test should be updated accordingly.

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+
+class FileWriterImplTest {
+    private final FileWriterImpl fileWriterImpl = new FileWriterImpl();
+
+    @Test
+    void write_validPath_Ok() throws IOException {
```
mateuszwojtkowiak: This test does not fully verify the behavior of the `write` method.  
If the file already exists, the test may pass even if the method does nothing.  
Consider deleting the file before the test or writing to a temporary file  
to ensure the test validates the actual write operation.

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+        String fileContent = Files.readString(Paths.get(path));
+        assertEquals(data, fileContent);
+    }
+
+    @Test
+    void write_invalidPath_NotOk() {
+        String path = "invalid/path/file.csv";
+        String data = "fruit,quantity\nbanana,100";
+
+        assertThrows(RuntimeException.class, () ->
```
mateuszwojtkowiak: The test expects a generic `RuntimeException`, but the implementation  
should throw a more specific exception type (e.g. `IllegalArgumentException`).  
The test should be updated to expect the correct exception type.

src/test/java/core/basesyntax/operation/BalanceOperationTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class BalanceOperationTest {
```
mateuszwojtkowiak, line 11: This test class covers only the happy path.  
According to the task requirements, BalanceOperation should also be tested  
for invalid inputs such as null transaction, null storage, negative quantity,  
and existing fruit replacement.

src/test/java/core/basesyntax/operation/PurchaseOperationTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class PurchaseOperationTest {
```
mateuszwojtkowiak, line 11: This test class covers only the happy path.  
According to the task requirements, PurchaseOperation must also be tested  
for invalid scenarios such as insufficient stock, missing fruit in storage,  
null transaction, null storage, and negative quantity.

src/test/java/core/basesyntax/operation/ReturnOperationTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class ReturnOperationTest {
```
mateuszwojtkowiak, line 11: This test class covers only the happy path.  
According to the task requirements, ReturnOperation must also be tested  
for invalid scenarios such as null transaction, null storage, negative quantity,  
and returning a fruit that does not yet exist in storage.

src/test/java/core/basesyntax/operation/SupplyOperationTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class SupplyOperationTest {
```
mateuszwojtkowiak, line 11: This test class covers only the happy path.  
According to the task requirements, SupplyOperation must also be tested  
for invalid scenarios such as null transaction, null storage, negative quantity,  
and supplying a fruit that does not yet exist in storage.

src/test/java/core/basesyntax/report/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+        Map<String, Integer> fruits = Map.of("banana", 100, "apple", 200);
+        for (var entry : fruits.entrySet()) {
+            storage.add(entry.getKey(), entry.getValue());
+        }
+
+        ReportGenerator reportGenerator = new ReportGeneratorImpl(storage);
+
+        String report = reportGenerator.getReport();
+
+        assertEquals("fruit,quantity\nbanana,100\napple,200\n", report);
```
mateuszwojtkowiak: The test assumes a specific ordering of fruits in the generated report.  
Since StorageImpl uses a HashMap, the iteration order is not guaranteed.  
The test should not rely on a fixed order or should use a deterministic map  
(e.g. LinkedHashMap) if ordering is required.

src/test/java/core/basesyntax/report/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax.report;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class ReportGeneratorImplTest {
```
mateuszwojtkowiak, line 10: The test class covers only basic scenarios.  
Additional edge cases should be tested, such as null storage,  
invalid fruit names, negative quantities, and zero quantities.

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+import core.basesyntax.strategy.OperationStrategy;
+import core.basesyntax.strategy.OperationStrategyImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class ShopServiceImplTest {
```
mateuszwojtkowiak, line 13: These tests are not isolated. ShopServiceImplTest currently executes real handler logic  
(ReturnOperation) and real StorageImpl behavior. According to the task requirements,  
ShopService should be tested independently, without testing strategy or handler logic.  
Handlers should be replaced with simple stubs that only record whether they were called.

mateuszwojtkowiak, line 13: The test class covers only basic scenarios.  
Additional edge cases should be tested: null transaction list, null elements inside the list,  
missing handler for an operation, and invalid transaction fields.

src/test/java/core/basesyntax/storage/StorageImplTest.java
```diff
@@ -0,0 +1,63 @@
+package core.basesyntax.storage;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class StorageImplTest {
+    @Test
+    void storageExist_ok() {
```
mateuszwojtkowiak: The test `storageExist_ok` does not verify any actual behavior.  
It only checks that two new StorageImpl instances are equal,  
which will always be true regardless of implementation correctness.  
This test should be removed or replaced with a meaningful scenario.

src/test/java/core/basesyntax/storage/StorageImplTest.java
```diff
@@ -0,0 +1,63 @@
+package core.basesyntax.storage;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class StorageImplTest {
```
mateuszwojtkowiak, line 8: The test class covers only basic scenarios and equality checks.  
Important edge cases are missing, such as:
- adding a fruit that does not exist yet,
- removing from a non-existing fruit,
- removing with exact quantity (should delete the key),
- getQuantity for a missing fruit,
- put() overriding existing values,
- negative quantities,
- null fruit names.
These cases should be added to properly test StorageImpl behavior.

mateuszwojtkowiak (changes requested):

src/test/java/core/basesyntax/report/ReportGeneratorImplTest.java
mate-student26: "invalid fruits names"
Do I need to specify the list of valid fruit names ? 
And then compare and validate every fruit added to the storage with that list ?

mate-student26: @mateuszwojtkowiak 
"invalid fruits names"
Do I need to specify the list of valid fruit names ?
And then compare and validate every fruit added to the storage with that list ?

mateuszwojtkowiak: You don't need to create a valid fruit list. Invalid means that you pass for example empty string "" or null.

src/test/java/core/basesyntax/storage/StorageImplTest.java
mate-student26: @mateuszwojtkowiak 
Adding 

`class FruitTransactionTest {
    @Test
    void negativeQuantity_notOk () {
        FruitTransaction fruitTransaction = new FruitTransaction();
        assertThrows(IllegalArgumentException.class, () ->
                fruitTransaction.setQuantity(-100));
    }
}`

relieves me of the necessity testing negative quantity in test StorageImpl ?

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
+package core.basesyntax.data;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class DataConverterImplTest {
```
mateuszwojtkowiak, line 10: The test suite includes a case for `null` input, but not for an empty list.  
An empty input should also be validated, as it represents a different scenario than `null`.  
Consider adding a test expecting `IllegalArgumentException` for an empty list.

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+        List<FruitTransaction> transactions = dataConverter
+                .convertToTransaction(inputReport);
+        assertEquals(2, transactions.size());
+    }
+
+    @Test
+    void convertToTransaction_nullInput_notOk() {
+        List<String> inputReport = null;
+        assertThrows(NullPointerException.class, () -> dataConverter
```
mateuszwojtkowiak: Using `NullPointerException` here is not ideal, because NPE usually indicates a programming error rather than invalid input provided by the caller.  
For cases where the method receives an incorrect argument (such as `null`), it's better to throw `IllegalArgumentException`.  
This makes the contract of the method clearer, improves debuggability, and explicitly communicates that the issue comes from invalid input rather than an internal failure.

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+    void convertToTransaction_validInput_ok() {
+        List<String> inputReport = List.of(
+                "fruit,quantity",
+                "b,banana,100",
+                "s,apple,50"
+        );
+
+        List<FruitTransaction> transactions = dataConverter
+                .convertToTransaction(inputReport);
+        assertEquals(2, transactions.size());
```
mateuszwojtkowiak, line 23: This test verifies only the size of the resulting list, but does not check whether the parsed transactions contain correct values.  
To ensure the converter works properly, consider asserting the operation type, fruit name, and quantity for each created `FruitTransaction`.

src/test/java/core/basesyntax/data/DataConverterImplTest.java
mateuszwojtkowiak, line 10: The converter should handle invalid numeric values, but there is no test verifying this behavior.  
Consider adding a test case where the quantity cannot be parsed (e.g., `"b,banana,abc"`), expecting `IllegalArgumentException`.

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    void convertToTransaction_nullInput_notOk() {
+        List<String> inputReport = null;
+        assertThrows(NullPointerException.class, () -> dataConverter
+                .convertToTransaction(inputReport));
+    }
+
+    @Test
+    void convertToTransaction_invalidLineFormat_notOk() {
+        List<String> inputReport = List.of(
```
mateuszwojtkowiak, line 55: This test fails because the second line has an invalid format, but it still does not verify that the header line is skipped.  
Even if the implementation incorrectly tries to parse the header, this test would still pass.  
Instead of creating a new test, you can update this one so that the data line is valid and the test checks that only the actual data (not the header) is processed.

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.data;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+
+public class FileReaderImplTest {
```
mateuszwojtkowiak, line 14: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+
+class FileWriterImplTest {
```
mateuszwojtkowiak, line 12: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/operation/BalanceOperationTest.java
mateuszwojtkowiak, line 11: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/operation/PurchaseOperationTest.java
mateuszwojtkowiak, line 11: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/operation/ReturnOperationTest.java
mateuszwojtkowiak, line 11: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/operation/SupplyOperationTest.java
mateuszwojtkowiak, line 11: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/report/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,32 @@
...
```
mateuszwojtkowiak, line 10: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+import core.basesyntax.strategy.OperationStrategyImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+class ShopServiceImplTest {
```
mateuszwojtkowiak, line 13: Please read my previous comments in this class and apply changes.

src/test/java/core/basesyntax/storage/StorageImplTest.java
mateuszwojtkowiak: No, testing negative quantity in FruitTransactionTest does not replace testing it in StorageImplTest.  
Each class has its own responsibility and must be validated independently.  
FruitTransaction validates user input, while StorageImpl manipulates stored quantities and may receive values from different sources.  
Therefore, StorageImpl must also be tested for negative quantities, null fruit names, and other edge cases.

src/test/java/core/basesyntax/storage/StorageImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test
+    void validLinesInStorage_ok() {
+        Storage expected = new StorageImpl();
+        expected.put("banana", 100);
+
+        Storage actual = new StorageImpl();
+        actual.put("banana", 100);
+        assertEquals(expected, actual);
```
mateuszwojtkowiak, line 17: This test compares two `StorageImpl` instances using `assertEquals`, but `StorageImpl` does not override `equals()` or `hashCode()`.  
As a result, the test does not actually verify the internal state of the storage.  
You should compare the underlying map returned by `getAll()` instead of comparing the objects directly.

src/test/java/core/basesyntax/storage/StorageImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    void removeWhenCurrentQuantityLessThanQuantity_notOk() {
```
mateuszwojtkowiak: The test checks removal with insufficient quantity, but it does not validate how the method behaves when the quantity argument itself is invalid (e.g., negative or zero).  
These cases should also be covered to ensure consistent input validation.

src/test/java/core/basesyntax/storage/StorageImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    void showAllInput_ok() {
+        Storage actual = new StorageImpl();
+        actual.put("banana", 100);
+
+        Storage expected = new StorageImpl();
+        expected.put("banana", 100);
+        assertEquals(expected.getAll(), actual.getAll());
```
mateuszwojtkowiak, line 50: This test does not verify any real behavior of `StorageImpl`.  
It manually constructs two separate storage instances with identical data and then compares their maps, which will always succeed regardless of the actual implementation.  
The test should instead validate how `StorageImpl` builds and updates its internal state through real operations, rather than comparing two manually prepared objects.

src/test/java/core/basesyntax/transactions/FruitTransactionTest.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax.transactions;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class FruitTransactionTest {
```
mateuszwojtkowiak, line 8: The current tests cover some invalid cases, but several important scenarios are still missing.  
You should also test:  
- blank fruit name (" ")  
- null operation  
- zero quantity  
- at least one valid (positive) transaction scenario  
Right now the class only tests negative paths, so the behavior for valid inputs is not verified.

src/test/java/core/basesyntax/transactions/FruitTransactionTest.java
```diff
@@ -0,0 +1,28 @@
...
+    void negativeQuantity_notOk() {
+        FruitTransaction fruitTransaction = new FruitTransaction();
+        assertThrows(IllegalArgumentException.class, () ->
+                fruitTransaction.setQuantity(-100));
+    }
+
+    @Test
+    void setFruit_nullFruit_notOk() {
+        FruitTransaction fruitTransaction = new FruitTransaction();
+        assertThrows(NullPointerException.class, () ->
```
mateuszwojtkowiak: Should be IllegalArgumentException.

mateuszwojtkowiak (changes requested):
Good job! Please review my comments and apply all the changes requested in my previous review.

src/test/java/core/basesyntax/operation/BalanceOperationTest.java
mate-student26: @mateuszwojtkowiak 
I have already test for negative quantity in FruitTransactionTest

@Test
    void setQuantity_negativeQuantity_notOk() {
        FruitTransaction fruitTransaction = new FruitTransaction();
        assertThrows(IllegalArgumentException.class, () ->
                fruitTransaction.setQuantity(-100));
    }
    
Shoudl I check the negative quantity also in every operation handler test, using Mockito ? 

Otherwise, test is going to fail because setting quantity in operation handler test throws exception :
at core.basesyntax.transactions.FruitTransaction.setQuantity(FruitTransaction.java:46)

mateuszwojtkowiak: You should still test the negative quantity case in each operation handler test.
The fact that FruitTransactionTest already checks the setter validation does not replace the need to verify how each handler behaves when it receives an invalid transaction.

Mockito is not the right tool here.
FruitTransaction is a data object, not a dependency, so it should not be mocked. Mocking DTOs creates artificial scenarios that never happen in real code and hides real bugs.

mate-student26: @mateuszwojtkowiak 
Thanks!
So, to sum up, I have to remove validation:

        if (quantity <= 0) {
            throw new IllegalArgumentException("Quantity must be greater than zero.");
        }

from setter 

        FruitTransaction setQuantity(int quantity)
        
And add validation to every handler. Is that correct?

mateuszwojtkowiak: You should not remove the validation from the setters in FruitTransaction, and you should not move this validation into every handler.
The validation inside the model is correct and should stay exactly where it is.

FruitTransaction is a data object (a DTO).
It is responsible for validating its own fields (fruit, quantity, operation).
This ensures that the object can never exist in an invalid state.

Handlers (BalanceOperation, SupplyOperation, etc.) are responsible only for business logic, not for validating the internal state of the DTO. Duplicating validation in every handler would break SRP, create code duplication, and make the model unsafe.

src/test/java/core/basesyntax/report/ReportGeneratorImplTest.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax.report;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import org.junit.jupiter.api.Test;
+
+class ReportGeneratorImplTest {
```
slade13, line 10: Currently only error cases are tested:

```
void getReport_nullStorage_notOk()
void getReport_emptyStorage_notOk()
```

You should add at least one positive test.

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax.data;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+public class FileReaderImplTest {
```
slade13, line 14: `FileReaderImplTest` tests only negative scenarios, add a test that reads an existing test file and checks the returned lines.

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,50 @@
+package core.basesyntax.data;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class FileWriterImplTest {
```
slade13, line 12: `FileWriterImplTest` tests only negative scenarios, add a test that writes data to a temporary file and verifies the file content.

src/test/java/resources_test/finalReport.csv
slade13, line 1: Test resources are placed in the wrong folder, the task says test files should be placed here:

```
src/test/resources
```

But the project uses:

```
src/test/java/resources_test
```

Move these files:

```
src/test/java/resources_test/finalReport.csv
src/test/java/resources_test/finalReport.txt
src/test/java/resources_test/reportToRead.csv
```

to:

```
src/test/resources
```

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+                "s,apple,50"
+        );
+
+        List<FruitTransaction> transactions = dataConverter
+                .convertToTransaction(inputReport);
+        assertEquals(2, transactions.size());
+    }
+
+    @Test
+    void convertToTransaction_emptyInput_notOk() {
```
slade13: This test is not testing an empty input list. It is testing an invalid empty line. You should either rename it:
```
convertToTransaction_emptyLine_notOk
```
or add a real empty list test.

src/main/java/core/basesyntax/transactions/FruitTransaction.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getFruit() {
+        return fruit;
+    }
+
+    public FruitTransaction setFruit(String fruit) {
+        if (fruit.contains(" ")) {
+            throw new IllegalArgumentException("Fruit contains spaces!");
+        }
+
+        if (fruit.equals(null)) {
```
slade13: Fix null checks in `FruitTransaction` - this code is incorrect:

```
if (fruit.contains(" ")) {
    throw new IllegalArgumentException("Fruit contains spaces!");
}

if (fruit.equals(null)) {
    throw new IllegalArgumentException("Fruit can't be null!");
}
```

If `fruit` is `null`, `fruit.contains(" ")` throws `NullPointerException` before your validation.

Use:

```
if (fruit == null) {
    throw new IllegalArgumentException("Fruit can't be null!");
}
if (fruit.isEmpty()) {
    throw new IllegalArgumentException("Fruit can't be empty!");
}
if (fruit.contains(" ")) {
    throw new IllegalArgumentException("Fruit contains spaces!");
}
```

The same issue exists here:

```
if (operation.equals(null)) {
    throw new IllegalArgumentException("Operation can't be null!");
}
```

Use:

```
if (operation == null) {
    throw new IllegalArgumentException("Operation can't be null!");
}
```

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+        shopService.process(transactions);
+        assertEquals(expectedStorage, storage);
+    }
+
+    @Test
+    void nullTranscationList_notOk() {
+        OperationStrategy operationStrategy = Mockito.mock(OperationStrategy.class);
+        Storage storage = Mockito.mock(Storage.class);
+        ShopService shopService = new ShopServiceImpl(operationStrategy, storage);
+        assertThrows(NullPointerException.class, () ->
```
slade13: Avoid expecting `NullPointerException `in tests. It is better to validate input explicitly in the production code and throw `IllegalArgumentException `with a clear message.

src/test/java/core/basesyntax/operation/SupplyOperationTest.java
```diff
@@ -0,0 +1,53 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.storage.Storage;
+import core.basesyntax.storage.StorageImpl;
+import core.basesyntax.transactions.FruitTransaction;
+import org.junit.jupiter.api.Test;
+
+class SupplyOperationTest {
+    private FruitTransaction fruitTransaction;
+    private ReturnOperation returnOperation;
```
slade13: Remove unused field - it is unused and also confusing because this is a test for `SupplyOperation`.

src/main/java/core/basesyntax/report/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+                throw new IllegalArgumentException("Null or empty keys are not allowed");
+            }
+
+            report.append(entry.getKey())
+                    .append(",")
+                    .append(entry.getValue())
+                    .append("\n");
+        }
+
+        return report.toString();*/
```
slade13: Remove commented-out code from production classes - commented-out code should not be left in the final solution.

```suggestion
```

slade13 (changes requested):
Please review my comments and add necessary changes.

Additionally to my comments on your source code, some test method names do not follow the required convention.

The required convention is:

```
<methodUnderTest>_<state>_<expectedBehavior>
```

Examples that should be renamed:

```
multipleTransactions_Ok()
emptyTransactionList_Ok()
missingHandler_notOk()
validLinesInStorage_ok()
addingNotExistingFruit_Ok()
showAllInput_ok()
```

Better:

```
process_validTransactions_ok()
process_emptyTransactionList_ok()
process_missingHandler_notOk()
put_validFruit_ok()
add_newFruit_ok()
getAll_validStorage_ok()
```

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    void process_nullTranscationList_notOk() {
+        OperationStrategy operationStrategy = Mockito.mock(OperationStrategy.class);
+        Storage storage = Mockito.mock(Storage.class);
+        ShopService shopService = new ShopServiceImpl(operationStrategy, storage);
+        assertThrows(IllegalArgumentException.class, () ->
+                shopService.process(null));
+    }
+
+    @Test
+    void process_missingHandler_notOk() {
```
slade13: There is no dedicated test class for `OperationStrategyImpl` and this test is placed here in ShopServiceImplTest.

This test checks the strategy, not `ShopServiceImpl`. Move it to:
```
src/test/java/core/basesyntax/strategy/OperationStrategyImplTest.java
```
Also add a positive strategy test that verifies that the appropriate handler is called.

src/test/java/core/basesyntax/operation/SupplyOperationTest.java
```diff
@@ -0,0 +1,52 @@
...
+                () -> supplyOperation.operate(fruitTransaction, null));
+    }
+
+    @Test
+    void operate_negativeQuantity_notOk() {
+        fruitTransaction = new FruitTransaction();
+        SupplyOperation supplyOperation = new SupplyOperation();
+        storage = new StorageImpl();
+        assertThrows(IllegalArgumentException.class,
+                () -> supplyOperation.operate(fruitTransaction.setQuantity(-1), storage));
```
slade13: Here method `operate()` is never called, exception is thrown by your setter class `setQuantity()`:
```
public FruitTransaction setQuantity(int quantity) {
        if (quantity < 0) {
            throw new IllegalArgumentException("Quantity can't be negative");
        }

        this.quantity = quantity;
        return this;
    }
```
So this test is testing your setter instead intended method. Since `FruitTransaction `does not allow a negative quantity, testing this scenario in every operation handler is redundant. It is already correctly tested in `FruitTransactionTest`.

Remove these tests from:
```
BalanceOperationTest
PurchaseOperationTest
ReturnOperationTest
SupplyOperationTest
```

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+    void write_emptyPath_notOk() {
+        String path = "";
+        String data = "fruit,quantity\nbanana,100";
+        assertThrows(IllegalArgumentException.class,
+                () -> fileWriterImpl.write(data, path));
+    }
+
+    @Test
+    void write_emptyData_notOk() {
+        String path = "src/test/java/resources_test/finalReport.csv";
```
slade13: Your test resources should be placed in:
```
src/test/resources
```

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+    void write_nullPath_notOk() {
+        String path = null;
+        String data = "fruit,quantity\nbanana,100";
+        assertThrows(IllegalArgumentException.class,
+                () -> fileWriterImpl.write(data, path));
+    }
+
+    @Test
+    void write_nullData_notOk() {
+        String path = "src/test/java/resources_test/finalReport.csv";
```
slade13: Your test resources should be placed in:
```
src/test/resources
```

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+                fileWriterImpl.write(data, path)
+        );
+    }
+
+    @Test
+    void write_validDataOutput_ok() {
+        Storage storage = new StorageImpl();
+        storage.put("banana", 100);
+        ReportGenerator report = new ReportGeneratorImpl(storage);
+        String path = "src/main/resources/finalReport.csv";
```
slade13: Tests should not read or modify production resources. This can modify the project state and make tests dependent on files used by `Main`. Test files should be located in:

```
src/test/resources
```

For example:

```
src/test/resources/reportToRead.csv
```

For writing tests, prefer `@TempDir`:

```
@TempDir
Path tempDirectory;

@Test
void write_validData_writesReport() throws IOException {
    Path output = tempDirectory.resolve("report.csv");
    String data = "fruit,quantity\nbanana,100";

    fileWriterImpl.write(data, output.toString());

    assertEquals(data, Files.readString(output));
}
```

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,40 @@
...
+        assertThrows(IllegalArgumentException.class, () ->
+                fileReaderImpl.read("src/test/java/resources_test/reportToRead.txt"));
+    }
+
+    @Test
+    void read_validAllData_ok() {
+        ArrayList<String> expectedData = new ArrayList<>();
+        expectedData.add("type,fruit,quantity");
+        expectedData.add("b,banana,100");
+        String path = "src/main/resources/reportToRead.csv";
```
slade13: Your test resources should be placed in:
```
src/test/resources
```

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+    void write_invalidPath_NotOk() {
+        String path = "invalid/path/file.csv";
+        String data = "fruit,quantity\nbanana,100";
+        assertThrows(IllegalArgumentException.class, () ->
+                fileWriterImpl.write(data, path)
+        );
+    }
+
+    @Test
+    void write_validDataOutput_ok() {
```
slade13: The positive writer test uses multiple unrelated components:

```
Storage storage = new StorageImpl();
ReportGenerator report = new ReportGeneratorImpl(storage);
FileWriter writer = new FileWriterImpl();
FileReader readFromFinalReport = new FileReaderImpl();
```

This is an integration test rather than an isolated unit test for `FileWriterImpl`.

When testing the writer:

-  provide a plain string directly; 
-  write it to a temporary file; 
-  verify the result using `Files.readString()`. 

Do not use `ReportGeneratorImpl` or `FileReaderImpl` in this test.

src/test/java/core/basesyntax/data/FileReaderImplTest.java
```diff
@@ -0,0 +1,40 @@
+package core.basesyntax.data;
+
+import static org.junit.Assert.assertEquals;
```
slade13: Do not use assertions from JUnit4, use this instead:
```
import static org.junit.jupiter.api.Assertions.assertEquals;
```

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void convertToTransaction_validInput_ok() {
+        List<String> inputReport = List.of(
+                "fruit,quantity",
+                "b,banana,100",
+                "s,apple,50"
+        );
+
+        List<FruitTransaction> transactions = dataConverter
+                .convertToTransaction(inputReport);
+        assertEquals(2, transactions.size());
```
slade13, line 23: This test still checks only the number of transactions:
```
assertEquals(2, transactions.size());
```
This does not prove that values were converted correctly.

Check all important fields:
```
    assertEquals(2, transactions.size());

    FruitTransaction first = transactions.get(0);
    assertEquals(FruitTransaction.Operation.BALANCE, first.getOperation());
    assertEquals("banana", first.getFruit());
    assertEquals(100, first.getQuantity());

    FruitTransaction second = transactions.get(1);
    assertEquals(FruitTransaction.Operation.SUPPLY, second.getOperation());
    assertEquals("apple", second.getFruit());
    assertEquals(50, second.getQuantity());
```

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        Mockito.verify(operationStrategy).process(transaction, storage);
+    }
+
+    @Test
+    void process_emptyTransactionList_ok() {
+        Storage expectedStorage = new StorageImpl();
+        Storage storage = new StorageImpl();
+        Map<FruitTransaction.Operation, OperationHandler> handlers = new HashMap<>();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(handlers);
```
slade13: This test creates a real strategy even though it is testing an empty transaction list:
```
Map<FruitTransaction.Operation, OperationHandler> handlers = new HashMap<>();
OperationStrategy operationStrategy = new OperationStrategyImpl(handlers);
```
Use mocks and verify that the strategy is not called:
```
@Test
void process_emptyTransactionList_doesNotCallStrategy() {
    OperationStrategy strategy = Mockito.mock(OperationStrategy.class);
    Storage storage = Mockito.mock(Storage.class);
    ShopService service = new ShopServiceImpl(strategy, storage);

    service.process(List.of());

    Mockito.verifyNoInteractions(strategy);
}
```

slade13 (changes requested):
Good job, please review my comments and fix these changes too:

1. Some names do not fully follow the requested convention or contain spelling/capitalization problems:
```
remove_WhenCurrentQuantityLessThanQuantity_notOk
remove_WhenFruitNotExist_notOk
remove_ExactQuantityRemoveKey_ok
put_OverrideExistingValue_ok
write_invalidPath_NotOk
```
Use consistent lower camel case after the first underscore:
```
remove_insufficientQuantity_notOk
remove_missingFruit_notOk
remove_exactQuantity_setsQuantityToZero
put_existingFruit_replacesQuantity
write_invalidPath_notOk
```

2. Do not commit generated coverage reports, project contains:
```
htmlReport/
```
Generated coverage reports should not normally be committed. Add them to .gitignore and remove the directory from the repository.

src/test/java/core/basesyntax/data/FileWriterImplTest.java
```diff
@@ -0,0 +1,66 @@
+package core.basesyntax.data;
+
+import static org.junit.Assert.assertEquals;
```
slade13: `FileWriterImplTest `still uses JUnit 4 assertion, use JUnit 5 instead:
```
import static org.junit.jupiter.api.Assertions.assertEquals;
```

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+    void process_nullTranscationList_notOk() {
+        OperationStrategy operationStrategy = Mockito.mock(OperationStrategy.class);
+        Storage storage = Mockito.mock(Storage.class);
+        ShopService shopService = new ShopServiceImpl(operationStrategy, storage);
+        assertThrows(IllegalArgumentException.class, () ->
+                shopService.process(null));
+    }
+
+    @Test
+    void process_missingHandler_notOk() {
```
slade13: This method does not test `ShopServiceImpl`. It tests `OperationStrategyImpl`, so it should be removed from `ShopServiceImplTest`.

You already have the correct place for this test in `OperationStrategyImplTest`.

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+    void process_emptyTransactionList_doesNotCallStrategy() {
+        OperationStrategy strategy = Mockito.mock(OperationStrategy.class);
+        Storage storage = Mockito.mock(Storage.class);
+        ShopService service = new ShopServiceImpl(strategy, storage);
+        service.process(List.of());
+        Mockito.verifyNoInteractions(strategy);
+    }
+
+    @Test
+    void process_nullTranscationList_notOk() {
```
slade13: Fix typo:

```suggestion
    void process_nullTransactionList_notOk() {
```

src/test/java/core/basesyntax/data/DataConverterImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+        List<String> inputReport = List.of(
+                "fruit, quantity",
+                ""
+        );
+        assertThrows(IllegalArgumentException.class, () -> dataConverter
+                .convertToTransaction(inputReport));
+    }
+
+    @Test
+    void convertToTransaction_nullInput_notOk() {
```
slade13, line 47: You test null input and invalid lines, but not an empty list, add:
```
@Test
void convertToTransaction_emptyInput_notOk() {
    assertThrows(IllegalArgumentException.class,
            () -> dataConverter.convertToTransaction(List.of()));
}
```

src/test/java/core/basesyntax/transactions/FruitTransactionTest.java
slade13, line 8: Currently you test:
```
setQuantity_negativeQuantity_notOk
setFruit_nullFruit_notOk
setFruit_emptyFruit_notOk
```
Add tests for the remaining validation paths:
```
setFruit_fruitWithSpaces_notOk
setOperation_nullOperation_notOk
setFruit_validFruit_ok
setQuantity_validQuantity_ok
setOperation_validOperation_ok
```

src/main/java/core/basesyntax/service/ShopServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public void process(List<FruitTransaction> transactions) {
+        if (transactions == null) {
+            throw new IllegalArgumentException("Transaction can't be null");
+        }
+
+        for (FruitTransaction transaction : transactions) {
+            if (transaction.getFruit() == null) {
```
slade13, line 28: Improve null transaction handling, if the list contains null, this will throw `NullPointerException`.

Add test to cover this case:
```
@Test
void process_transactionIsNull_notOk() {
    OperationStrategy operationStrategy = Mockito.mock(OperationStrategy.class);
    Storage storage = Mockito.mock(Storage.class);
    ShopService shopService = new ShopServiceImpl(operationStrategy, storage);

    assertThrows(IllegalArgumentException.class,
            () -> shopService.process(List.of((FruitTransaction) null)));
}
```

src/main/java/core/basesyntax/storage/StorageImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    public void put(String fruit, int quantity) {
+        if (fruit == null) {
+            throw new IllegalArgumentException("Fruit cannot be null");
+        }
+
+        transactions.put(fruit, quantity);
+    }
+
+    @Override
+    public void add(String fruit, int quantity) {
```
slade13, line 37: put() validates null fruit:
```
if (fruit == null) {
    throw new IllegalArgumentException("Fruit cannot be null");
}
```
But add() does not:
```
@Override
public void add(String fruit, int quantity) {
    int current = transactions.getOrDefault(fruit, 0);
    transactions.put(fruit, current + quantity);
}
```
This allows storing null as a key.

Add similar validation:
```
@Override
public void add(String fruit, int quantity) {
    if (fruit == null) {
        throw new IllegalArgumentException("Fruit cannot be null");
    }

    int current = transactions.getOrDefault(fruit, 0);
    transactions.put(fruit, current + quantity);
}
```
And add a test:
```
@Test
void add_nullFruit_notOk() {
    Storage storage = new StorageImpl();

    assertThrows(IllegalArgumentException.class,
            () -> storage.add(null, 100));
}
```

slade13 (changes requested):
Good job! There are are still some small issues or missing test cases. Please review my comments and add necessary changes.

src/test/java/core/basesyntax/service/ShopServiceImplTest.java
mate-student26: @slade13  how you have noted this :P

slade13 (approved):
Good job, there are still some small issues like some methods do not follow naming convention:
```
<methodUnderTest>_<state>_<expectedBehavior>
```
but mostly your code looks good and is compliant with the task description.

PR #1322, opened 2026-06-19
for check

acidden:


finalReport.csv
```diff
@@ -0,0 +1,3 @@
+fruit,quantity
```
Elena-Bruyako, line 1: for the future, files should be stored in `test/resources` folder

Elena-Bruyako (approved):
good job
see comment for the future

PR #1332, opened 2026-07-27
I did

Yehor0820:


Elena-Bruyako (changes requested):
write tests for read and write operations

use `@AfterEach` to `Storage.getFruits().clear();`

PR #1338, opened 2026-08-07
Task complete

Hubris0:
Prepared test coverage for creating transaction report for supplied transactions.


src/test/java/core/basesyntax/MainTest.java
```diff
@@ -0,0 +1,174 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class MainTest {
```
slade13: Tests should be separated by responsibility, currently everything is placed inside:
```
class MainTest
```
This is directly against the task requirements. You should create separate test classes for the components you want to verify, for example:
```
ReaderImplTest
WriterImplTest
ReportCreatorTest

BalanceTransactionTest
SupplyTransactionTest
PurchaseTransactionTest
ReturnTransactionTest

OperationStrategyImplTest
TransactionHandlerTest
```
Creating `MainTest `still suggests that you're testing the application's complete flow rather than individual units.

slade13 (changes requested):
You have already written several useful scenarios and you are thinking about invalid input, insufficient fruit quantity, and the complete application flow, which is a good start. However, the most important requirement of this task is currently missing: **unit tests must be isolated**.

Right now almost the whole application is tested through one `MainTest`, with multiple services, strategies, file operations, and handlers participating in the same tests. Split these tests into separate classes and test every component directly. Prepare `Storage` directly when testing transactions or report creation instead of reaching the tested functionality through other services.

Also improve test naming so that the first part contains the actual method under test, remove assertions that only verify the manually prepared input, and move test files to `src/test/resources`.

slade13 (changes requested):
Please review my comments, additionally fields like:
```
private static final String fruitName
private static final int fruitQuantity
```
are constants and should normally use uppercase naming:
```
private static final String FRUIT_NAME = "testfruit";
private static final int FRUIT_QUANTITY = 123;
```

Please fix things mentioned by my comments and request another review.

mateuszwojtkowiak (changes requested):

slade13 (changes requested):
Good job, your code is much closer to proper unit testing and the major structural problems from the previous submissions have been fixed.

The main remaining improvement is isolation: `WriterImplTest `should not use `ReaderImpl`, and `TransactionHandlerTest `doesn't need to retest all four transaction implementations because those already have dedicated tests.

Please review my comments and update your code.

mateuszwojtkowiak (approved):
Great job!

